PR #36, opened 2025-10-24
Support Webpack-based frameworks, require Storybook 10.1

JReinhold:
This PR migrates to using the new `experimental_devServer` preset property introduced in https://github.com/storybookjs/storybook/pull/32816 . It also means that it now requires Storybook versions that has that property, which currently are only canaries.

I manually tested that this worked both with Vite and Webpack, by modifying the Storybook in this reop.

### DIFF
--- a/packages/addon-mcp/README.md
+++ b/packages/addon-mcp/README.md
@@ -15,7 +15,7 @@ The addon provides tools to improve agents' UI development capabilities, retriev
 ### Installation and Setup
 
 > [!NOTE]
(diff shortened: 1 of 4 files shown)


packages/addon-mcp/README.md
```diff
@@ -18,7 +15,7 @@ The addon provides tools to improve agents' UI development capabilities, retriev
 ### Installation and Setup
 
 > [!NOTE]
-> This addon requires Storybook 9.0 or higher.
+> This addon requires Storybook 10.1 or higher, currently only available as the canary verison `0.0.0-pr-32816-sha-28f59f53`.
```
JReinhold: we'll update this version string everywhere whenever the PR to core has been merged and released.

PR #42, opened 2025-10-27
Add optional handlers to `@storybook/mcp` and telemetry to the new tools in `@storybook/addon-mcp`

JReinhold:
- You can now add `onListAllComponents` and `onGetComponentDocumentation` when creating the MCP handler or when calling it with a request. These handlers will be called with information about the manifests.
- Similarly there is now also an optional `onSessionInitialize` handler that can be passed when creating the MCP handler (but not passed with any request).
- `addon-mcp` uses this to collect telemetry on these two calls, similar to the existing tools.


kasperpeulen (commented):
LGTM

PR #44, opened 2025-10-28
Support Storybook 9.1.16 and up

JReinhold:
Given the `experimental_devServer` API has been released in Storybook 9.1.16 and 10.0.0 via https://github.com/storybookjs/storybook/pull/32862, we can relax the version requirements here.

I've confirmed that the addon works in Storybook 9.1.16


shilman:
Nice job documenting all of this clearly @JReinhold !! ❤️

PR #48, opened 2025-10-28
Configure toolsets via addon options and headers

JReinhold:
Added the ability to enable/disable "toolsets", either the Core Tools or the Component Documentation Tools.

They can be configured via addon options, and both default to true.

They can also be configured via the `X-MCP-Toolsets` header, that users can set when setting up the MCP server. this configuration API directly mimics [how you can similarly configure the remote GitHub MCP](https://github.com/github/github-mcp-server/blob/main/docs/remote-server.md#optional-headers). I'm not married to this specific structure of the API, it was just easy to copy what they had done.

I'm also open to different naming of the toolsets instead of `core` and `componentDocumentation`. Instead of `core` I also thought about `storiesDevelopment`, `storybookDevelopment`, `storybookContext`...

See the updated readme for a more detailed explanation of configuring toolsets.

Component Documentation Tools are still conditioned on manifests being present, no matter how much you try to enable them via toolset configuration.

### DIFF
--- a/.changeset/tame-deer-draw.md
+++ b/.changeset/tame-deer-draw.md
@@ -2,4 +2,4 @@
 '@storybook/addon-mcp': patch
 ---
 
-Add possiblity to configure toolsets (dev tools vs docs tools) either via addon options or request headers
+Add possibility to configure toolsets (dev tools vs docs tools) either via addon options or request headers


shilman (commented):
WDYT about the following structure:

```
- dev:
  - get-story-urls
  - get-ui-building-instructions
- docs:
  ...
- test:
   ...
```

PR #50, opened 2025-10-29
GET handler serving HTML, updated manifest format

JReinhold:
This PR adds a GET handler to `/mcp` in `@storybook/addon-mcp` that shows the addon is successfully running. When component manifests exists, it automatically redirects to human readable component manifests. See https://github.com/storybookjs/storybook/pull/32882

Recording of the experience of visiting `/mcp` now:

https://github.com/user-attachments/assets/2285d215-32df-4f5e-a2ef-1b26b22d7a96

It also updates the manifest format to account for errors and missing snippets. See https://github.com/storybookjs/storybook/pull/32855

### DIFF
--- a/.changeset/nine-jars-send.md
+++ b/.changeset/nine-jars-send.md
@@ -1,5 +1,0 @@
----
-'@storybook/mcp': patch
----
-
-Add optional "enabled" function to the directly exported tool adders. This allow you to define a function that will dynamically enable/disable the tool however you want, eg. per request condition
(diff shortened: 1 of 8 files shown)


shilman (approved):
I imagine we'll want to redesign this but this is a great first iteration.

Longer term I worry that people might not know what the component manifest is and so it might be better to have a list of toolsets and statuses, rather than a redirect:

```
Dev [description] [disabled]
Docs [description] [view manifest]
```

kasperpeulen (approved):

PR #51, opened 2025-10-29
chore: update `tmcp`

paoloricciuti:
Updated tmcp together with the transport...since this version is technically a breaking, updating both of them at the same time makes the breaking transparent!

Sorry for the breaking, hope this PR makes you forgive me! 🧡


JReinhold (approved):
Changesets? 😘
https://tenor.com/view/where-you-at-gif-21177622

PR #106, opened 2025-12-05
Eval: Use Commander for args parsing

JReinhold:
Telescoping on to #104

This replaces the args collection logic with Commander.

Previously we were using a combination of Node's `parseArgs`, Valibot transformation+validation, manual env var retrieved and custom --help logic. These are all replaced with Commander, with some custom env var logic still.

The clack-based prompts remain the same.

In terms of functionality, the only difference is that the `--upload / --no-upload / --upload-id XYZ` option+env var has been simplified to just `--upload-id XYZ / --no-upload-id`.

### DIFF
--- a/eval/lib/collect-args.ts
+++ b/eval/lib/collect-args.ts
@@ -43,7 +43,6 @@ function parseBooleanEnv(
 	}
 
 	const normalized = envVar.toLowerCase().trim();
-	console.log({ normalized });
 	if (['true', '1', 'yes'].includes(normalized)) {
 		return true;
 	}


eval/lib/collect-args.ts
```diff
@@ -16,180 +16,253 @@ export type CollectedArgs = {
...
+	const program = new Command()
+		.name('eval.ts')
+		.description(
+			'A CLI tool for testing AI coding agents with Storybook and MCP tools.',
+		)
+		.argument('[eval-name]', 'Name of the eval directory in evals/')
+		.addOption(
+			new Option('-a, --agent <name>', 'Which coding agent to use')
+				.choices(['claude-code'])
+				.env('AGENT'),
```
valentinpalkovic: I didn't know about the `.env` method. Pretty interesting!

valentinpalkovic (approved):

PR #120, opened 2026-01-05
Support MDX docs from manifests

JReinhold:
Companion PR in core: https://github.com/storybookjs/storybook/pull/33408

This PR adds support for the new docs entries in manifests, that addon-docs now adds based on MDX files. It's not required, so it will still support Storybooks that don't use addon-docs.

See the following tests for how this is outputted to the LLM:

- https://github.com/storybookjs/mcp/pull/120/changes#diff-5fab611b4f17ecb443641b29c27bfdfa29bd8d358b7817a8e3622b8213b1a93cR360-R379
- https://github.com/storybookjs/mcp/pull/120/changes#diff-f24efa15324100f21ba9d2e690e8f6478198d7b926368e1f7e6fc4299c8a960dR262-R276

# Breaking Changes 

This PR introduces a number of minor breaking changes to `@storybook/mcp`:

1. The lower level tool adder functions have been renamed:
    1. `addGetComponentDocumentationTool` -> `addGetDocumentationTool`
    2. `addListAllComponentsTool` -> `addListAllDocumentationTool`
2. The optional tool hooks have been renamed:
    1. `onListAllComponents` -> `onListAllDocumentation`
    2. `onGetComponentDocumentation` -> `onGetDocumentation`
3. The exported `MANIFEST_PATH` constant have been removed in favor of two new constants, `COMPONENT_MANIFEST_PATH` and `DOCS_MANIFEST_PATH`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,14 +7,14 @@ settings:
 catalogs:
   default:
     '@storybook/addon-a11y':
-      specifier: 10.2.0-alpha.12
-      version: 10.2.0-alpha.12
+      specifier: 10.2.0-alpha.14
+      version: 10.2.0-alpha.14
     '@storybook/addon-docs':
-      specifier: 10.2.0-alpha.12
-      version: 10.2.0-alpha.12
+      specifier: 10.2.0-alpha.14
+      version: 10.2.0-alpha.14
     '@storybook/react-vite':
-      specifier: 10.2.0-alpha.12
-      version: 10.2.0-alpha.12
+      specifier: 10.2.0-alpha.14
+      version: 10.2.0-alpha.14
     '@tmcp/adapter-valibot':
       specifier: ^0.1.4
       version: 0.1.4
@@ -25,8 +25,8 @@ catalogs:
       specifier: ^0.4.1
       version: 0.4.1
     storybook:
-      specifier: 10.2.0-alpha.12
-      version: 10.2.0-alpha.12
+      specifier: 10.2.0-alpha.14
+      version: 10.2.0-alpha.14
     tmcp:
       specifier: ^1.16.0
       version: 1.18.1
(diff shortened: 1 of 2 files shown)


packages/mcp/src/utils/manifest-formatter/markdown.ts
```diff
@@ -108,13 +108,21 @@ export const markdownFormatter: ManifestFormatter = {
 		return parts.join('\n').trim();
 	},
 
-	formatComponentManifestMapToList(manifest: ComponentManifestMap): string {
+	formatDocsManifest(doc) {
+		return dedent`# ${doc.title}
+
+			${doc.content}`;
+	},
```
JReinhold: The first line here might be unnecessary/redundant, adding the title? We could also just assume that MDX files always has a `# title`?

.changeset/large-carrots-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@storybook/addon-mcp': patch
+---
+
+Add support for docs entries in manifests, sourced by MDX files.
```
valentinpalkovic: Is this needed?

.changeset/stale-ducks-pay.md
```diff
@@ -0,0 +1,17 @@
+---
```
valentinpalkovic: Is this needed? if so, please adjust the file name and adjust the ordered list

eval/lib/collect-args.ts
```diff
@@ -681,14 +681,15 @@ export async function collectArgs(): Promise<CollectedArgs> {
 	return result;
 }
 
+// TODO: add support for docs manifests too
```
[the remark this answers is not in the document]
valentinpalkovic: Hmm. Do we have to do a prompting at all? If the user selects "Storybook MCP Docs", shouldn't it just load the `components.json` and the `docs.json` file? Is there a scenario where you need to select?

valentinpalkovic (approved):
LGTM. Left some small comments. Please clean up the pnpm-lock.yaml file. (Remove the experiments)

PR #123, opened 2026-01-08
Added a new tool `get-component-story-documentation` for retrieving specific story documentation

valentinpalkovic:
This PR splits the `get-documentation` tool into two focused tools to significantly reduce token usage while maintaining or even improving coverage (more research/proof needed). The main tool now returns a concise summary, and a new on-demand tool fetches specific story implementations when needed.

### Tool Split

1. **`get-documentation`**: Now returns a summary format with:
   - Import statement
   - Props table (TypeScript prop definitions)
   - First 3 primary usage examples with code snippets
   - List of remaining story names (without code snippets)

2. **`get-documentation-for-story`** (new): Fetches detailed documentation for a specific story variant on demand. Agents can use this when they need to see the implementation of a specific story that wasn't included in the summary.

## Results

Based on initial benchmarks (N=1, further benchmarks needed):

- **Cost**: $0.5967 → $0.3872 (**~35% reduction**)
- **Token count**: 29,962 → 13,686 (**~54% reduction**)
- **Coverage**: 96% → **100%** ✅

The improved coverage suggests that agents are now more effectively using the focused summary format and fetching specific stories only when needed, rather than being overwhelmed by too much information upfront.

## Technical Details

- `MAX_STORIES_TO_SHOW` constant set to 3 (configurable in `packages/mcp/src/utils/manifest-formatter/types.ts`)
- Summary format implemented in both markdown and XML formatters
- New `formatStoryDocumentation()` method added to `ManifestFormatter` interface
- Tool descriptions updated to guide agents on when to use each tool

### DIFF
--- a/.changeset/eighty-snakes-judge.md
+++ b/.changeset/eighty-snakes-judge.md
@@ -1,0 +1,7 @@
+---
+'@storybook/mcp': patch
+---
+
+Minimize token usage by only including the 3 first stories in component documentation.
+
+... if there are already prop types. If there are no prop types, include all stories. Additional stories can be fetched individually using a new `get-documentation-for-story` tool.


packages/mcp/src/utils/manifest-formatter/markdown.ts
```diff
@@ -154,4 +174,36 @@ export const markdownFormatter: ManifestFormatter = {
 
 		return parts.join('\n').trim();
 	},
+
+	formatStoryDocumentation(componentManifest, storyName) {
```
JReinhold: this must be able to be reused for the `formatComponentManifest`-function above, which essentially does the same thing.

PR #131, opened 2026-01-20
Add testing toolset

JReinhold:
Works on https://github.com/storybookjs/mcp/issues/138

Companion PR: https://github.com/storybookjs/storybook/pull/33206

This PR introduces a testing toolset to `addon-mcp`. The toolset includes a `run-story-tests` tool that, when called, uses the existing `addon-vitest` configuration to run tests and report results back to the agent. If `addon-a11y` is configured, a11y violations will be reported to the agent. This can be disabled by an `a11y: false` input to the tool, based on the user's instructions to the agent.

## Eval changes

I've added extensive eval to ensure this feature works, see eval tasks 908-912 and their readmes.

- Added eval tasks 908-912
- rename internal dev evals from 111-117 to 901-907 to indicate that 9xx are internal evals
- Added variant config that contains all permutations of the testing toolset, to easily run them together in multiple iterations.
- To support this, I've added an `--internal` flag to the simple `eval.ts` setup, which must be used to show internal variant configs. This is to not bloat the eval experience for outsiders
- Eval 912 tests that a11y violations are handled correctly. Specifically, we want to ensure that the agent doesn't auto-fix visual violations; instead, it asks the user how to proceed. To grade this, I've added an LLM-as-a-Judge grader to the eval harness, which will take any evidence provided and a scoring guideline and score it appropriately, with a rationale. It's very flexible and supports using any file/content produced in the trial as evidence. You can see an example of this in action here: https://6916512dc7d0d3267c8cc898-dvubtzsova.chromatic.com/?path=/docs/judge--docs
-

### DIFF
--- a/packages/addon-mcp/src/tools/run-story-tests.ts
+++ b/packages/addon-mcp/src/tools/run-story-tests.ts
@@ -388,7 +388,7 @@ function formatRunStoryTestResults({
 	}
 
 	const a11yReports = testResults.a11yReports as Record<StoryId, A11yReport[]>;
-	const a11yViolationCount = countA11yViolations(a11yReports);
+	const a11yViolationCount = runA11y ? countA11yViolations(a11yReports) : 0;
 	if (runA11y && a11yReports && Object.keys(a11yReports).length > 0) {
 		const a11ySection = formatA11yReportsSection({ a11yReports, origin });
 		if (a11ySection) {


packages/mcp/src/types.ts
```diff
@@ -99,7 +98,7 @@ export const ComponentManifest = v.object({
 	import: v.optional(v.string()),
 	stories: v.optional(v.array(Story)),
 	// loose schema for react-docgen types, as they are pretty complex
-	reactDocgen: v.optional(v.custom<Documentation>(() => true)),
+	reactDocgen: v.optional(v.any()),
```
JReinhold: This single line change fixed a type bundling bug, that rolldown caught after the upgrade. Here's a summary:

Image

packages/addon-mcp/src/tools/run-story-tests.ts
```diff
@@ -0,0 +1,325 @@
...
+				const passingStories = componentTestStatuses.filter(
+					(status) => status.value === 'status-value:success',
+				);
+				const failingStories = componentTestStatuses.filter(
+					(status) => status.value === 'status-value:error',
+				);
+
+				if (passingStories.length > 0) {
+					sections.push(
+						`## Passing Stories\n\n- ${passingStories.map((status) => status.storyId).join('\n- ')}`,
```
JReinhold: don't use \n in this file but actual newlines in the template string

packages/addon-mcp/src/tools/tool-names.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Tool name constants extracted to avoid circular dependencies.
+ *
+ * The cycle was:
+ * - types.ts → get-storybook-story-instructions.ts (for GET_UI_BUILDING_INSTRUCTIONS_TOOL_NAME)
+ * - get-storybook-story-instructions.ts → run-story-tests.ts (for RUN_STORY_TESTS_TOOL_NAME)
+ * - run-story-tests.ts → types.ts (for StoryInputArray)
```
JReinhold:
```suggestion
```

packages/addon-mcp/src/tools/run-story-tests.ts
```diff
@@ -0,0 +1,325 @@
...
+				const textResult = sections.join('\n\n');
+
+				return {
+					content: [
+						{
+							type: 'text',
+							text: textResult,
+						},
+					],
+				};
```
JReinhold: we need to add telemetry calls here similar to what we do in the the other tools. ideally it should log similar inputs, and counts of the output (pass/fail/a11y violations/unhandled errors), or maybe more if relevant.

pnpm-workspace.yaml
```diff
@@ -6,29 +6,32 @@ packages:
...
 catalog:
-  '@storybook/addon-a11y': 10.3.0-alpha.4
-  '@storybook/addon-docs': 10.3.0-alpha.4
-  '@storybook/addon-vitest': 10.3.0-alpha.4
-  '@storybook/react-vite': 10.3.0-alpha.4
+  '@storybook/addon-a11y': 0.0.0-pr-33206-sha-06900f2a
+  '@storybook/addon-docs': 0.0.0-pr-33206-sha-06900f2a
+  '@storybook/addon-vitest': 0.0.0-pr-33206-sha-06900f2a
+  '@storybook/react-vite': 0.0.0-pr-33206-sha-06900f2a
+  '@storybook/addon-themes': 0.0.0-pr-33206-sha-06900f2a
```
JReinhold: will be reverted when PR has been released in the monorepo

packages/addon-mcp/src/utils/find-story-ids.test.ts
```diff
@@ -0,0 +1,191 @@
...
+				title: 'Input',
+				importPath: './src/Input.stories.tsx',
+				tags: ['story'],
+			},
+		},
+	};
+
+	beforeEach(() => {
+		vi.clearAllMocks();
+	});
```
JReinhold: Applied. I removed the redundant `vi.clearAllMocks()` setup from this test file (and the now-unused `beforeEach` import). Targeted tests still pass.

packages/addon-mcp/src/tools/run-story-tests.ts
```diff
@@ -0,0 +1,325 @@
...
+	{ a11yEnabled }: { a11yEnabled: boolean },
+) {
+	const addonVitestConstants = await getAddonVitestConstants();
+	const testRunQueue = createAsyncQueue();
+
+	const description =
+		`Run tests for one or more stories and report accessibility issues. For visual/design accessibility violations (for example color contrast), ask the user before changing styles.
+Use this continously to monitor test results as you work on your UI components and stories.
+Results will include passing/failing status` +
+		(a11yEnabled ? ', and accessibility violation reports.' : '');
```
JReinhold: Fixed — updated the typo from "continously" to "continuously" in the tool description.

packages/addon-mcp/src/tools/run-story-tests.ts
```diff
@@ -0,0 +1,325 @@
...
+				case 'cancelled':
+					reject(new Error('Test run was cancelled'));
+					break;
+				default:
+					reject(new Error('Unexpected test run response'));
+			}
+		};
+
+		channel.on(triggerTestRunResponseEventName, handleResponse);
+
```
JReinhold: Applied. `triggerTestRun()` now has a per-run timeout and centralized cleanup so listeners are always removed and the promise always settles (success/error/cancel/timeout/emit failure). Added focused timeout + listener cleanup tests.

JReinhold (commented):
Addressed all current review threads on this PR: applied targeted fixes for valid comments, left informational-only thread as no-op, and added/updated focused tests where behavior changed. Local formatting and targeted suites pass; full check run only fails on the unrelated dependency-version gate in internal-storybook.

kasperpeulen (approved):
LGTM

JReinhold:
This PR is now ready to go, it's just waiting for the companion PR in core to get released, so we don't have to depend on canary releases here.

PR #156, opened 2026-02-13
Add E2E code coverage

kasperpeulen:
Collect V8 coverage from the Storybook process spawned during E2E tests and upload it to Codecov alongside unit test coverage.

- Enable source maps in tsdown builds so V8 coverage can map back to TypeScript source
- Set `NODE_V8_COVERAGE` in `test:ci` to collect coverage from the spawned Storybook process
- Add `test:ci:report-e2e-coverage` script using `c8` with `monocart-coverage-reports` to convert V8 coverage to lcov
- Use monocart instead of c8's default v8-to-istanbul to produce line counts that align with vitest's AST-based coverage (within 1-3 lines per file), enabling proper Codecov merging
- Filter lcov output to only include `src/` files, matching vitest's `include` pattern
- Upload both unit test and E2E coverage lcov files to Codecov
- Add `.npmignore` to both packages to exclude `.map` files from published packages

Coverage: 73.90% → 74.96% (+1.06%)


package.json
```diff
@@ -24,7 +24,8 @@
 		"release": "turbo run build && pnpm changeset publish",
 		"storybook": "turbo watch storybook --filter=@storybook/mcp-internal-storybook",
 		"test": "vitest",
-		"test:ci": "vitest run --coverage --reporter=default --reporter=github-actions --reporter=junit --outputFile=test-report.junit.xml",
+		"test:ci": "NODE_V8_COVERAGE=./e2e-coverage vitest run --coverage --reporter=default --reporter=github-actions --reporter=junit --outputFile=test-report.junit.xml",
```
AriPerkkio, line 27: `NODE_v8_COVERAGE` can be replaced with `vitest --coverage.trackProcessAndWorker` in future:

- https://github.com/vitest-dev/vitest/pull/9976

PR #160, opened 2026-02-17
Fix bug where attached docs were not part of the component documentation

JReinhold:
This PR fixes an unfortunate regression, where MDX docs attached to components were not being output by the MCP server, even though they were present in the manifest.


kasperpeulen (approved):
LGTM

PR #163, opened 2026-02-18
fix(addon-mcp): preview-stories toolset cant find stories in monorepo packages

domazet93:
### Context
`preview-stories` returns `"story not found"` for stories that live outside the Storybook app directory, even though they render fine in Storybook.

In a monorepo like this:
```
my-monorepo/
  apps/storybook/          <- Storybook runs here
  packages/design-system/  <- stories live here
```
The tool computes a path with a `./` prefix, but `http://localhost:6006/index.json` doesnt include it for paths that go up to parent directories:
```
tool builds: `./../../packages/design-system/Button.stories.tsx`
index.json importPath has: `../../packages/design-system/Button.stories.tsx`
```
https://github.com/storybookjs/mcp/blob/main/packages/addon-mcp/src/tools/preview-stories.ts#L126
strict === comparison fails which results in "story not found"

### Fix
Strip `./` from both sides before comparing.
Added 2 test cases + fixture for monorepo path resolution.


.changeset/fix-monorepo-import-path.md
```diff
@@ -0,0 +1,17 @@
...
+
+```
+my-monorepo/
+  apps/storybook/          ← Storybook runs here
+  packages/design-system/  ← stories live here
+```
+
+The tool builds `./../../packages/design-system/Button.stories.tsx` but `index.json` stores `../../packages/design-system/Button.stories.tsx` (no `./`). The strict `===` fails silently.
+
+Path normalization now uses `normalizeStoryPath` from `storybook/internal/common` to stay consistent with how Storybook core canonicalizes `importPath`.
```
JReinhold: for the changelog we don't need this much information, but it's great that it is in the PR description!
```suggestion
```

PR #175, opened 2026-02-25
Add support for using story ID as input for preview and test tools

JReinhold:
This pull request introduces story ID-based inputs to the preview and testing tools, making it easier for agents and tools to reference stories directly.

* Added story ID-based input shapes for `preview-stories` and `run-story-tests`, while retaining path-based story inputs.
* Included story IDs in documentation outputs (`list-all-documentation` and `get-documentation`), allowing agents to discover and reuse IDs without extra filesystem lookup steps.
* New evaluation tasks are added to ensure agents use the most context-appropriate input strategy.

Previously, we saw numerous times that the agent would use the docs tools to get information about a component, then traverse the file system to find the story files so it could get story IDs for previewing/testing stories. That was inefficient. Now story IDs can be used directly, so if the agent doesn't have the file already in context, it will use the docs tool to get the IDs, and then preview/test stories.
In situations where it already has the file in context, it will still used the path-based inputs.

Evals shows that this is both faster and cheaper (in the minimal scenario).

### DIFF
--- a/eval/tasks/915-preview-story-by-id/README.md
+++ b/eval/tasks/915-preview-story-by-id/README.md
@@ -16,10 +16,10 @@ Asks the agent to preview two existing stories: Button Primary and Secondary.
 
 ## Quality Signal
 
-| Metric                                                                                                                      | Weight |
-| --------------------------------------------------------------------------------------------------------------------------- | ------ |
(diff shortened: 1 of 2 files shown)


kasperpeulen (approved):
LGTM

PR #226, opened 2026-05-18
feat: mcp-proxy package

huang-julien:
https://github.com/storybookjs/storybook/issues/34826

This PR provides the `@storybook/mcp-proxy` package to be used by ADE plugins.

The MCP communicates through stdio.

## Tools proxying

  It proxies all tools available in `@storybook/mcp`. Hard-coded ,no need to make it dynamic, as `@storybook/mcp-proxy` releases will be kept in sync with`@storybook/mcp`.

## Tool calls

When a tool is being called, `mcp-proxy` requires a `cwd` argument: the absolute path of the Storybook project the call should target. It must match exactly the cwd   from which `storybook dev` was started./

If no Storybook is registered at that cwd, the call returns a repair message (and, when other Storybooks are running, lists their cwds as candidates) instead of proxying.

## Instance registry

Each running `storybook dev` registers itself in a filesystem registry (PID, cwd, endpoint). The proxy reads this registry on every call to resolve the target
Storybook. Stale entries (whose PID is no longer alive) are pruned automatically on read, so a crashed Storybook doesn't leak into routing.

Instance records use a versioned schema (`v1`, valibot) so the registry format can evolve without breaking older proxies.

## Transport

- Stdio for the proxy ↔ MCP client (ADE) link
- HTTP/SSE for the proxy ↔ Storybook link

## Usage

Run pnpm --filter @storybook/mcp-proxy inspect to explore tools via the MCP Inspector.

## Manual testsing 

`pnpm --filter=@storybook/mcp-proxy run-storybook <absolute path to a porject with SB` runs a storybook dev command and creates a record instance file. We can remove this once SB 10.5.0 is released. This is jsut for tessting out the behavior. 

## Open questions

- Should Storybook itself expose the Record Instance Schema ? Or export types ? Or should `mcp-proxy` have it's own record instance schema ?


kasperpeulen (approved):
LGTM

Maybe it could use one more e2e like test?
I still see some references in the code to the 227 branch, maybe you can clean that up?
And maybe you can improve the README.md a bit before merging!

PR #233, opened 2026-05-21
Addon MCP: Make endpoint configurable

kasperpeulen:
## TL;DR

Adds an optional `endpoint` option to `@storybook/addon-mcp` so the addon dev server can serve MCP somewhere other than `/mcp`.

Default behavior is unchanged: without `endpoint`, the addon still registers `/mcp`. The option only configures the local addon route; composed remote refs and OAuth metadata keep the existing `/mcp` behavior.

Companion PR: storybookjs/storybook#34863.

## Usage

```ts
// .storybook/main.ts
export default {
  addons: [
    {
      name: '@storybook/addon-mcp',
      options: { endpoint: '/tools/mcp' },
    },
  ],
};
```

## Review Map

Start with the behavior surface:

1. `packages/addon-mcp/src/types.ts` parses `endpoint?: string` as a URL pathname.
2. `packages/addon-mcp/src/preset.ts` applies the parsed endpoint to the addon GET/POST MCP routes.
3. `packages/mcp/src/utils/get-manifest.ts` keeps Storybook manifests top-level by deriving `/manifests/...` from the request origin, independent of the MCP endpoint path.

Tests to review next:

- `packages/addon-mcp/src/types.test.ts` covers endpoint parsing and rejection directly at the schema boundary.
- `packages/addon-mcp/src/preset.test.ts` covers default `/mcp`, custom endpoints, no-refs behavior, and refs/local-manifest behavior.
- `packages/mcp/src/utils/get-manifest.test.ts` covers manifest URL resolution for default, trailing slash, nested custom endpoint, prefixed path, and localhost custom endpoint requests.
- `packages/addon-mcp/src/auth/composition-auth.test.ts` only adds `stubGlobal` cleanup; `composition-auth.ts` itself is not changed in the final diff.

Docs/release notes:

- `packages/addon-mcp/README.md` documents addon configuration.
- `README.md` mentions the configured endpoint in local debug instructions.
- `.changeset/quiet-rivers-connect.md` marks the new addon option as a minor change.

Mechanical/CI-required:

- `pnpm-workspace.yaml` and generated lockfiles refresh Storybook packages from `10.5.0-alpha.1` to `10.5.0-alpha.2`; the repo dependency freshness test requires this.

## Non-goals

- No runtime metadata preset is added in this PR.
- No remote-ref custom endpoint contract is added.
- No OAuth protected-resource metadata behavior is changed.
- Manifest files are not moved under the MCP endpoint. They remain Storybook top-level `/manifests/...` resources.

## Review Notes

This branch is now a single focused commit. Review the final Files changed diff in the order above: option schema, route wiring, shared manifest URL derivation, tests, docs, then generated lockfile/catalog refreshes.

## Validation

Local checks run on this branch:

```sh
pnpm vitest run --project=@storybook/mcp packages/mcp/src/utils/get-manifest.test.ts
pnpm vitest run --project=@storybook/addon-mcp packages/addon-mcp/src/preset.test.ts packages/addon-mcp/src/types.test.ts packages/addon-mcp/src/auth/composition-auth.test.ts
pnpm --filter @storybook/addon-mcp typecheck
pnpm format:check
pnpm lint
git diff --check
```

`pnpm lint` has 0 errors and the same 2 unrelated warnings in untouched test files.

### DIFF
--- a/packages/addon-mcp/src/types.ts
+++ b/packages/addon-mcp/src/types.ts
@@ -3,7 +3,22 @@ import type { Options } from 'storybook/internal/types';
 import { GET_TOOL_NAME, LIST_TOOL_NAME, type StorybookContext } from '@storybook/mcp';
 import { GET_UI_BUILDING_INSTRUCTIONS_TOOL_NAME } from './tools/tool-names.ts';
 
+const isLiteralEndpointPathname = (endpoint: string) => {
+	try {
+		const { pathname } = new URL(endpoint, 'http://storybook.local');
+		return pathname === endpoint && pathname !== '/';
+	} catch {
+		return false;
+	}
+};
+
 export const AddonOptions = v.object({
+	endpoint: v.optional(
+		v.pipe(
+			v.string(),
+			v.check(isLiteralEndpointPathname, 'Endpoint must be a literal URL pathname'),
+		),
+	),
 	toolsets: v.optional(
 		v.object({
 			dev: v.exactOptional(v.boolean(), true),
--- a/.changeset/quiet-rivers-connect.md
+++ b/.changeset/quiet-rivers-connect.md
@@ -1,0 +1,5 @@
+---
+"@storybook/addon-mcp": minor
+---
+
+Add an optional MCP endpoint setting for the addon dev server.
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ You can start Storybook with:
 pnpm storybook
 ```
 
-This will build everything and start up Storybook with addon-mcp, and you can then connect your coding agent to it at `http://localhost:6006/mcp` and try it out.
+This will build everything and start up Storybook with addon-mcp, and you can then connect your coding agent to it at `http://localhost:6006/mcp` (or your configured addon endpoint) and try it out.
 
 ### Working with the MCP App
 
--- a/apps/internal-storybook/pnpm-lock.yaml
+++ b/apps/internal-storybook/pnpm-lock.yaml
@@ -7,20 +7,20 @@ settings:
 catalogs:
   default:
     '@storybook/addon-a11y':
-      specifier: 10.5.0-alpha.1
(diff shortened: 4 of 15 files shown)


packages/addon-mcp/src/auth/composition-auth.ts
```diff
@@ -117,10 +117,10 @@ export class CompositionAuth {
 	}
 
 	/** Build .well-known/oauth-protected-resource response. */
-	buildWellKnown(origin: string): object | null {
+	buildWellKnown({ origin, endpoint }: { origin: string; endpoint: string }): object | null {
 		if (!this.#authRequirement) return null;
 		return {
-			resource: `${origin}/mcp`,
+			resource: `${origin}${endpoint}`,
```
huang-julien: Should we use `join` from `path` or `pathe` ? It makes it sasfer

packages/addon-mcp/src/types.ts
```diff
@@ -4,6 +4,7 @@ import { GET_TOOL_NAME, LIST_TOOL_NAME, type StorybookContext } from '@storybook
 import { GET_UI_BUILDING_INSTRUCTIONS_TOOL_NAME } from './tools/tool-names.ts';
 
 export const AddonOptions = v.object({
+	endpoint: v.optional(v.string()),
```
huang-julien:
```suggestion
	endpoint: v.optional(
  v.pipe(
    v.string(),
    v.startsWith('/', 'Endpoint must start with "/"'),
  ),
),
```

huang-julien (changes requested):

packages/addon-mcp/src/auth/composition-auth.ts
kasperpeulen: I guess we should use new URL because origin is an url not a path!

huang-julien (approved):

PR #235, opened 2026-05-21
Handle private composed Storybooks in MCP proxy

kasperpeulen:
## TL;DR

The local Storybook MCP proxy cannot authenticate into private composed Storybooks. This PR keeps those sources visible and returns normal MCP guidance telling the agent to use the composed Storybook's own `/mcp` endpoint, while direct non-proxy requests keep the existing OAuth challenge behavior.

Targets `main` directly.

## Behavior Contract

- Regular direct requests to `/mcp` still get the OAuth `401` challenge when composed refs require auth.
- Requests from `@storybook/mcp-proxy` are marked with `X-Storybook-MCP-Proxy: true`, so `addon-mcp` lets verified local proxy traffic reach the docs tools instead of challenging at the route boundary.
- The direct OAuth-client path and local-proxy path are represented once as request access mode, then baked into the per-request manifest provider.
- Composed sources are tracked as per-source state (`public`, `requires-auth`, or `unknown`) rather than parallel auth/source collections.
- Private composed refs remain listed in multi-source docs output. They are not filtered out.
- Private-source guidance is modeled as data via `ManifestProviderResult` / `SourceManifestFailure`, not a normal throw/catch path.
- `SourceManifestFailure` stays pure data (`kind`, `endpoint`, `authProvider`, or fetch-failure message); user-facing Markdown is formatted in `@storybook/mcp`.
- The docs tools are registered when local manifests exist or composed remote sources are available, including remote-only setups without local manifests.
- If auth is discovered during manifest fetch rather than startup probing, proxy-originated requests still return the same private-source guidance instead of turning into a 401/error path.

## Review Order

1. Proxy marker and route auth: `packages/mcp-proxy/src/utils/proxy-client.ts` adds the request header; `packages/addon-mcp/src/preset.ts` only bypasses the top-level OAuth challenge for verified local proxy requests.
2. Access/state model: `packages/addon-mcp/src/auth/composition-auth.ts` owns per-source auth state and creates a per-request manifest provider from explicit request access mode.
3. Source failure contract: `packages/mcp/src/types.ts` and `packages/mcp/src/utils/get-manifest.ts` model source failures as result data via `ManifestProviderResult` / `getManifestResult`.
4. Tool rendering: `packages/mcp/src/tools/*documentation*.ts` and `packages/mcp/src/utils/manifest-formatter/*` render source-failure data as normal tool content/list entries.
5. Remote docs registration: `packages/addon-mcp/src/mcp-handler.ts` and `packages/addon-mcp/src/preset.ts` enable docs tools/status for composed remote sources even without local manifests.
6. Test decomposition: `packages/addon-mcp/src/auth/composition-auth.fetch.test.ts` splits fetch/cache behavior out of the auth-state tests.

## Test Plan

- `pnpm exec oxfmt --check packages/addon-mcp/src/auth/composition-auth.ts packages/addon-mcp/src/auth/composition-auth.test.ts packages/addon-mcp/src/auth/composition-auth.fetch.test.ts packages/addon-mcp/src/auth/index.ts packages/addon-mcp/src/mcp-handler.ts packages/addon-mcp/src/mcp-handler.test.ts packages/addon-mcp/src/preset.ts packages/addon-mcp/src/preset.test.ts packages/mcp/src/types.ts packages/mcp/src/index.ts packages/mcp/src/utils/get-manifest.ts packages/mcp/src/tools/get-documentation.ts packages/mcp/src/tools/get-documentation-for-story.ts packages/mcp/src/tools/list-all-documentation.ts packages/mcp/src/tools/get-documentation.test.ts packages/mcp/src/tools/get-documentation-for-story.test.ts packages/mcp/src/tools/list-all-documentation.test.ts packages/mcp/src/utils/get-manifest.test.ts packages/mcp/README.md`
- `git diff --check`
- `pnpm turbo run typecheck --filter=@storybook/addon-mcp --filter=@storybook/mcp`
- `pnpm vitest --project=@storybook/mcp --run`
- `pnpm vitest --project=@storybook/addon-mcp --run`

### DIFF
--- a/apps/internal-storybook/tests/mcp-composition-auth.e2e.test.ts
+++ b/apps/internal-storybook/tests/mcp-composition-auth.e2e.test.ts
@@ -12,7 +12,7 @@ import {
 const PORT = 6008;
 const MCP_ENDPOINT = `http://localhost:${PORT}/mcp`;
 const WELL_KNOWN_ENDPOINT = `http://localhost:${PORT}/.well-known/oauth-protected-resource`;
-const STARTUP_TIMEOUT = 30_000;
+const STARTUP_TIMEOUT = 60_000;
 
 let storybookProcess: ReturnType<typeof x> | null = null;
 
--- a/packages/addon-mcp/src/auth/composition-auth.ts
+++ b/packages/addon-mcp/src/auth/composition-auth.ts
@@ -6,7 +6,13 @@
  * MCP clients like VS Code to handle the OAuth flow with Chromatic.
  */
 
-import { ComponentManifestMap, DocsManifestMap, type Source } from '@storybook/mcp';
+import {
+	ComponentManifestMap,
+	DocsManifestMap,
+	type ManifestProviderResult,
+	type ManifestSourceNotice,
+	type Source,
+} from '@storybook/mcp';
 import * as v from 'valibot';
 
 export interface ComposedRef {
@@ -14,6 +20,8 @@ export interface ComposedRef {
 	title: string;
 	url: string;
 }
+
+type RemoteSource = Source & { url: string };
 
 const OAuthResourceMetadata = v.object({
 	resource: v.optional(v.string()),
@@ -36,14 +44,21 @@ interface AuthRequirement {
 	serverMetadata: OAuthServerMetadata;
 }
 
+interface McpAuthCheck {
+	unauthorized: boolean;
+	authRequirement: AuthRequirement | null;
+}
+
 export type ManifestProvider = (
 	request: Request | undefined,
 	path: string,
 	source?: Source,
-) => Promise<string>;
+) => Promise<ManifestProviderResult>;
 
 const MANIFEST_CACHE_TTL = 60 * 60 * 1000; // 60 minutes
 const REVALIDATION_TTL = 60 * 1000; // 60 seconds
+export const STORYBOOK_MCP_PROXY_HEADER = 'X-Storybook-MCP-Proxy';
+export const STORYBOOK_MCP_PROXY_HEADER_VALUE = 'true';
 
 interface CacheEntry {
 	text: string;
@@ -52,7 +67,10 @@ interface CacheEntry {
 }
 
 export class AuthenticationError extends Error {
-	constructor(url: string) {
+	constructor(
+		url: string,
+		public readonly authRequirement: AuthRequirement | null = null,
+	) {
 		super(`Authentication failed for ${url}. Your token may be invalid or expired.`);
 		this.name = 'AuthenticationError';
 	}
@@ -61,7 +79,7 @@ export class AuthenticationError extends Error {
 export class CompositionAuth {
 	#authRequirement: AuthRequirement | null = null;
 	#authRequiredUrls: string[] = [];
-	#refsWithManifests: ComposedRef[] = [];
+	#sourceRefs: ComposedRef[] = [];
 	#manifestCache = new Map<string, CacheEntry>();
 	#lastToken: string | null = null;
 	#authErrors = new WeakMap<Request, AuthenticationError>();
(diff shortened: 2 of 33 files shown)


apps/internal-storybook/tests/mcp-composition-auth.e2e.test.ts
```diff
@@ -12,7 +12,7 @@ import {
 const PORT = 6008;
 const MCP_ENDPOINT = `http://localhost:${PORT}/mcp`;
 const WELL_KNOWN_ENDPOINT = `http://localhost:${PORT}/.well-known/oauth-protected-resource`;
-const STARTUP_TIMEOUT = 30_000;
+const STARTUP_TIMEOUT = 60_000;
```
huang-julien, line 15: Is this change because of native fetch default timeout being 60s ?

packages/addon-mcp/src/auth/composition-auth.ts
```diff
@@ -344,6 +366,24 @@ export class CompositionAuth {
...
+	return {
+		listText: `This composed Storybook is private and requires Chromatic authentication. Use its MCP endpoint: ${mcpEndpoint}`,
+		detailText: `# ${source.title}
+
+This composed Storybook is private and requires Chromatic authentication.
+
+To access documentation from this source, register or use its MCP endpoint:
+${mcpEndpoint}`,
+	};
+}
```
huang-julien: `AuthenticationError` can be caused by `isMcpUnauthorized` but also by any 401 response from manifest fetch. 

Do you think we should make this agnostic of chromatic ?

huang-julien (approved):

apps/internal-storybook/tests/mcp-composition-auth.e2e.test.ts
kasperpeulen: this is because my WIFI is really slow ;P

packages/addon-mcp/src/auth/composition-auth.ts
kasperpeulen: Yep, good point. I updated this so the notice only says Chromatic authentication for chromatic.com Storybook URLs. Other authenticated composed refs now get generic wording: the source requires authentication, and the agent should use the shown MCP endpoint if it has access.

apps/internal-storybook/tests/mcp-composition-auth.e2e.test.ts
kasperpeulen: Not because of native fetch. This timeout is just giving the composed Storybook startup/registry path more room in slow local/CI conditions; the test is waiting for the dev server + registry flow rather than relying on fetch timeout behavior.

packages/mcp-proxy/src/utils/proxy-client.ts
```diff
@@ -4,14 +4,17 @@ import type {
 	StorybookInstanceRecordV1,
 } from '../types/index.ts';
 
+const STORYBOOK_MCP_PROXY_HEADER = 'X-Storybook-MCP-Proxy';
+const STORYBOOK_MCP_PROXY_HEADER_VALUE = 'true';
+
```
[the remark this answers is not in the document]
kasperpeulen: Good callout. I am going to leave this duplicated for now rather than introduce a package edge just for one internal marker: `@storybook/mcp-proxy` should not depend on `@storybook/addon-mcp`, and adding a new shared package for this single header feels heavier than the risk. Both sides have focused coverage for the exact header/value, so we should catch drift in tests.

packages/mcp/src/utils/get-manifest.ts
```diff
@@ -146,7 +163,7 @@ export async function getManifests(
 		);
 	}
 
-	if (docsResult.status === 'rejected') {
+	if (docsResult.status === 'rejected' || isManifestSourceNotice(docsResult.value)) {
 		return { componentManifest };
 	}
```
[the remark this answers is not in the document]
kasperpeulen: I think the current behavior is the one we want here, and I added test coverage to make that explicit. A source-level private notice still wins when the component manifest path returns the notice, which is what the proxy/private-ref flow does. But if the component manifest is fetchable and only the optional docs manifest returns a notice, we keep the fetched component documentation rather than replacing useful docs with a notice.

packages/addon-mcp/src/preset.ts
```diff
@@ -71,8 +80,12 @@ export const experimental_devServer: PresetPropertyFn<'experimental_devServer'>
...
 		res: import('node:http').ServerResponse,
 	): boolean => {
-		const token = extractBearerToken(req.headers['authorization']);
-		if (compositionAuth.requiresAuth && !token) {
+		const token = extractBearerToken(req.headers.authorization);
+		if (
+			compositionAuth.requiresAuth &&
+			!token &&
+			!isStorybookMcpProxyRequest(req.headers['x-storybook-mcp-proxy'])
+		) {
```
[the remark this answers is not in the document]
kasperpeulen: Fixed. The proxy marker is no longer enough on its own: the auth bypass now only applies when the request also comes from a loopback address. A spoofed non-loopback request with `X-Storybook-MCP-Proxy: true` still receives the normal 401 challenge, and that path is covered in `preset.test.ts`.

PR #241, opened 2026-05-22
feat: tell agent to upgrade old storybook first

huang-julien:
https://github.com/storybookjs/storybook/issues/34826

Before proxying a call, we can check the cwd's Storybook verison using `createRequire` and looking at the package.json of reachable storybook.

If unsupported by storybook/addon-mcp, we can tell the agent to ask the user to upgrade storybook first through the upgrade skill.

### DIFF
--- a/packages/mcp-proxy/src/tools/intercepts.test.ts
+++ b/packages/mcp-proxy/src/tools/intercepts.test.ts
@@ -8,70 +8,66 @@ describe('intercepts', () => {
 		['mcp-starting', 'starting up'],
 		['mcp-error', 'reported an error'],
 		['invalid-cwd', 'absolute path'],
+		['storybook-too-old', 'storybook-upgrade'],
 	] as const)('%s contains an actionable hint', (reason, needle) => {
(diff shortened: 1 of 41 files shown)


packages/mcp-proxy/src/tools/intercepts.test.ts
```diff
@@ -8,48 +8,66 @@ describe('intercepts', () => {
 		['mcp-starting', 'starting up'],
 		['mcp-error', 'reported an error'],
 		['invalid-cwd', 'absolute path'],
+		['storybook-too-old', 'storybook-upgrade'],
 	] as const)('%s contains an actionable hint', (reason, needle) => {
 		expect(getInterceptMarkdown(reason)).toContain(needle);
 	});
 
+	it('storybook-too-old reports the detected version, the required version, and points to the upgrade skill', () => {
+		const md = getInterceptMarkdown('storybook-too-old', { version: '9.0.5' });
```
kasperpeulen, line 17: should be storybook 10.5 (or alpha)

PR #249, opened 2026-05-29
Implement get-stories-by-component tool

yannbf:
This PR adds a new MCP tool, `get-stories-by-component`, that answers the question: "Which stories render this component?"

You give it one or more component source file paths, and it returns the real Storybook story IDs that depend on those files. Under the hood it uses Storybook's live dependency graph, so the results are always accurate and up to date (including hot-reloaded changes during dev).

Results are ranked by distance — how directly a story uses the component:

0 = the path you passed is itself a story file
1 = a story for a component that imports it directly
2+ = a story that pulls it in transitively (e.g. through a page or wrapper)
This lets an AI agent take anything — a file it just edited, a feature the user named, a shared token/util — turn it into a list of component paths, and get back grounded story IDs it can hand to preview-stories (for preview URLs) or `display-review` (for a review page). It never has to guess IDs from filenames.


ghengeveld (commented):
A few non-blocking documentation/contract observations on the new `get-stories-by-component` tool, focused on token efficiency and keeping the tool's self-description consistent with its implementation and the server instructions.

packages/addon-mcp/src/tools/get-stories-by-component.ts
```diff
@@ -0,0 +1,305 @@
...
+
+Returns sorted results from the Storybook index — if a component has no matches here, it likely has no stories yet (say so, don't fabricate).
+
+Backed by Storybook's live reverse dependency graph: distance is the import-graph hop count from the story file to the component (1 = directly imported, 2+ = transitively). Available when the Storybook dev server is running with a builder that supports change detection (e.g. Vite); otherwise the tool returns a typed error.
+
+Results are sorted by \`distance\` (lower = stronger signal). Prefer the lowest-distance results first; widen only when needed. For shared components like Button or Icon, expect many indirect (\`distance\` ≥ 2) matches — pass \`maxDistance\` to cap noise.`,
```
ghengeveld: **Tool description efficiency / duplication.** This description is loaded into the model's context on *every* turn (it's part of the tools manifest), so length here is a recurring per-session token cost — quite different from per-call output. A large chunk of this prose also duplicates the new "Mapping any input to story IDs" section in `dev-instructions.md`, so the same guidance now lives in three places (this description, the schema field descriptions below, and the server instructions) and will inevitably drift.

Suggestion: keep the description to *what it does + when to reach for it vs `get-changed-stories` (one line) + the one hard constraint (never invent IDs; no matches = no stories yet)*, and move the workflow/strategy prose (the multi-paragraph decision tree and the "whenever the user describes a part of the UI…" paragraph) into the server instructions, which is the right home for it. That's roughly a 50% trim with no loss of grounding.

packages/addon-mcp/src/tools/get-stories-by-component.ts
```diff
@@ -0,0 +1,305 @@
...
+	storyId: v.string(),
+	title: v.string(),
+	name: v.string(),
+	importPath: v.string(),
+	distance: v.pipe(
+		v.number(),
+		v.description(
+			'Import-graph depth from the story file to the component (lower = stronger). 1: story file directly imports the component. 2+: reached through N hops.',
+		),
+	),
```
ghengeveld, line 52: **Self-documentation inconsistency for `distance`.** The field is described inconsistently across (and even within) surfaces:

- Here it reads "**1**: story file directly imports the component" with no mention of 0 — but the `componentPaths` description just above (line 27) says story files "appear at **distance 0** as self-matches," and `dev-instructions.md` documents "**0** = colocated, 1 = direct importer." The `serializeComponentSection` tests also construct matches at distance 0.
- Separately, `maxDistance` defaults to **3** in code (`DEFAULT_MAX_DISTANCE`, and the schema on line 36 says "Defaults to 3"), while `dev-instructions.md` step 3 tells the agent to "Default to `maxDistance: 2`."

Net effect: the agent sees two different definitions of `distance` and two different recommended defaults. Worth settling on one `distance` convention (0- or 1-based) and one default, then aligning this field, the tool description, the schema, the instructions, and the tests.

packages/addon-mcp/src/tools/get-stories-by-component.ts
```diff
@@ -0,0 +1,305 @@
...
+const GetStoriesByComponentInput = v.object({
+	componentPaths: v.pipe(
+		v.array(v.string()),
+		v.minLength(1),
+		v.description(
+			`Absolute paths to component source files (e.g. "/repo/src/Button.tsx").
+Pass the components you actually want stories for — typically files you just read, edited, or that the user mentioned.
+Story files (\`*.stories.*\`) are accepted too: they appear at distance 0 as self-matches, plus any reverse-graph hits (other stories that import them).`,
+		),
+	),
```
ghengeveld, line 30: **Input contract mismatch.** The schema states "Absolute paths to component source files," but `resolveComponentStories` calls `path.resolve(workingDir, componentPath)` and transparently accepts relative paths (there's even a test asserting relative paths resolve against the working dir). So the validated contract is looser than the documented one: an agent that follows the docs strictly sends only absolute paths, while one that reads the implementation can send relative — inconsistent behavior depending on which source it trusts.

Either tighten the schema to reject non-absolute inputs (e.g. a `v.check(path.isAbsolute, …)` on each entry), or relax the wording to say relative paths are accepted and resolved against the Storybook working directory.

packages/addon-mcp/src/tools/get-stories-by-component.ts
```diff
@@ -0,0 +1,305 @@
...
+
+Returns sorted results from the Storybook index — if a component has no matches here, it likely has no stories yet (say so, don't fabricate).
+
+Backed by Storybook's live reverse dependency graph: distance is the import-graph hop count from the story file to the component (1 = directly imported, 2+ = transitively). Available when the Storybook dev server is running with a builder that supports change detection (e.g. Vite); otherwise the tool returns a typed error.
+
+Results are sorted by \`distance\` (lower = stronger signal). Prefer the lowest-distance results first; widen only when needed. For shared components like Button or Icon, expect many indirect (\`distance\` ≥ 2) matches — pass \`maxDistance\` to cap noise.`,
+			schema: GetStoriesByComponentInput,
+			outputSchema: GetStoriesByComponentOutput,
+			enabled: () => server.ctx.custom?.toolsets?.dev ?? true,
```
ghengeveld, line 212: **Optional / soft suggestion: tool annotations.** `tmcp`'s tool options accept `annotations` (`ToolAnnotations`), and this tool is read-only and idempotent. Adding `annotations: { readOnlyHint: true, idempotentHint: true }` would let MCP clients reason about safety / auto-approval and is aligned with MCP guidance. None of the addon's tools currently set annotations, so this could be a small follow-up applied across all the read tools rather than just here — not blocking.

packages/addon-mcp/src/tools/get-stories-by-component.ts
yannbf: Thanks, done!

packages/addon-mcp/src/tools/get-stories-by-component.ts
yannbf: thanks, was already fixed in a previous commit

packages/addon-mcp/src/tools/get-stories-by-component.ts
yannbf: This seems like a good followup for another PR (to do in all tools) 👍

packages/addon-mcp/src/tools/get-stories-by-component.ts
yannbf: thanks, was already fixed in a previous commit

PR #251, opened 2026-05-29
fix(mcp-proxy): only trust localhost for mcp proxy

huang-julien:
https://github.com/storybookjs/mcp/pull/248#pullrequestreview-4379335829

@kasperpeulen need discussion

### DIFF
--- a/packages/mcp-proxy/src/utils/registry.test.ts
+++ b/packages/mcp-proxy/src/utils/registry.test.ts
@@ -68,6 +68,60 @@ describe('readRegistry', () => {
 		expect(result).toEqual([alive]);
 	});
 
+	it('drops records whose mcp.endpoint is not loopback', async () => {
+		const base = {
+			schemaVersion: 1,
+			pid: process.pid,
+			cwd: '/tmp/x',
+			url: 'http://localhost:6006',
+			port: 6006,
+		};
+		await fs.writeFile(
+			join(dir, 'remote.json'),
+			JSON.stringify({
+				...base,
+				instanceId: 'remote',
+				mcp: { status: 'ready', endpoint: 'http://attacker.test/mcp' },
+			}),
+		);
+		await fs.writeFile(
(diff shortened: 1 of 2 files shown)


kasperpeulen (changes requested):
Thanks for picking this up. The loopback check is a good hardening step, but I don't think it fully closes the issue that came up on the temp-dir registry change.

The concern there is not only "could the endpoint be remote?" It is also "can we trust this registry record at all?" On platforms where `tmpdir()` ends up in a shared location like `/tmp`, a different local user/process may be able to create a record that points to its own loopback MCP server, for example `http://127.0.0.1:49152/mcp`. This new check would accept that, because the endpoint is still loopback. So this blocks direct remote endpoints, but it does not prove that the record was created by the current user's Storybook process.

I think we should either make the registry location/provenance safe first, or be explicit that this PR is only partial defense-in-depth. The stronger fix would be something like: prefer a per-user registry dir (`XDG_RUNTIME_DIR` on Linux, or another uid-scoped temp path), and/or validate directory/file ownership and permissions before records become routable. The loopback check still belongs, but as the final guard after the record itself is trusted.

There is also a smaller design issue in the shape of the helper. `hasTrustedEndpoint(undefined) === true` reads strangely: a missing endpoint is not really trusted, it is just valid for non-ready records. And the helper validates `new URL(endpoint)`, while the fetch path later resolves `new URL(endpoint, record.url)`. That gives us two endpoint models to keep in sync.

Can we collapse this into one boundary helper that validates the parsed record as a whole? For example: if the record is ready, require an endpoint, resolve the actual target URL the proxy will fetch, check `http:` plus loopback host, and return the record or drop it. For starting/not-installed/error states, allow the missing endpoint explicitly by status rather than by treating `undefined` as trusted.

Requesting changes because I don't want us to merge this as "trusted endpoint" protection while the registry provenance question is still open.

PR #256, opened 2026-06-02
[Storybook 11] - BREAKING: Resolve the story index in-process via presets

yannbf:
Replace the loopback `fetch(${origin}/index.json)` with an in-process lookup through `options.presets.apply('storyIndexGenerator').getIndex()`. The dev server registers and memoises a single StoryIndexGenerator, so this returns the live, HMR-fresh index with no network round-trip.

- add utils/get-story-index.ts (getStoryIndex(options)) + tests
- migrate get-changed-stories, preview-stories and run-story-tests from fetchStoryIndex(origin) to getStoryIndex(options)
- remove utils/fetch-story-index.ts and update test mocks


packages/addon-mcp/src/utils/get-story-index.ts
JReinhold, line 1: IMO this whole function could be removed, if you instead of passing the `options` around, would just pass the `() => getIndex()` function around in the server context.

So in the addon's preset, you would resolve the `storyIndexGenerator` and create a function like `const getIndex = () => storyIndexGenerator.getIndex()` that would be added to the server context.

But it's not a strong opinion, just a way to simplify the code a little.

JReinhold (changes requested):
You have to bump the peer dependency range of Storybook packages to 10.2.0 and above, as that's where the story index generator was introduced: https://github.com/storybookjs/storybook/releases/tag/v10.2.0-alpha.9

Via a minor version changeset.

ghengeveld (approved):
Besides Jeppe's versioning concern, LGTM.

PR #265, opened 2026-06-02
Add dynamic tools to Available Toolsets

yannbf:
This adds dynamic tools which were not available in the template, as it was only adding static tools.

image

### DIFF
--- a/packages/addon-mcp/src/mcp-handler.ts
+++ b/packages/addon-mcp/src/mcp-handler.ts
@@ -20,14 +20,13 @@ import { collectTelemetry } from './telemetry.ts';
 import type { AddonContext, AddonOptionsOutput } from './types.ts';
 import { logger } from 'storybook/internal/node-logger';
 import { getManifestStatus } from './tools/is-manifest-available.ts';
-import { getReviewStatus } from './utils/is-review-available.ts';
+import { getToolAvailability } from './utils/get-tool-availability.ts';
 import { addRunStoryTestsTool, getAddonVitestConstants } from './tools/run-story-tests.ts';
 import { estimateTokens } from './utils/estimate-tokens.ts';
 import { isAddonA11yEnabled } from './utils/is-addon-a11y-enabled.ts';
 import type { CompositionAuth } from './auth/index.ts';
 import { buildServerInstructions } from './instructions/build-server-instructions.ts';
 import { DEFAULT_MCP_ENDPOINT } from './constants.ts';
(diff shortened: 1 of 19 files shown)


packages/addon-mcp/src/utils/get-tool-availability.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Single source of truth for the runtime gates that decide whether the
+ * change-detection and review tools are registered.
+ *
+ * Used both by the MCP server (to actually register the tools) and by the
+ * browser landing page (to show accurate enabled/disabled badges), so the two
+ * can never drift apart.
+ */
+export async function getToolAvailability(
```
JReinhold, line 45: this reads like a common "get the availability of all dynamic tools", but it's missing the test tool (requires addon-vitest), the a11y sub-feature of the test tool, and the docs tool (depends on manifest enablement).

those today exists as individual functions/booleans, but they should either be part of this, or everything should be individual. It doesn't make sense that some tools are calculated in here, and some are calculated individually.

JReinhold (changes requested):

JReinhold (approved):

packages/addon-mcp/src/mcp-handler.ts
```diff
@@ -61,11 +59,11 @@ const initializeMCPServer = async (options: Options, multiSource?: boolean) => {
...
 			return buildServerInstructions({
 				devEnabled: server?.ctx.custom?.toolsets?.dev ?? true,
-				testEnabled: (server?.ctx.custom?.toolsets?.test ?? true) && !!addonVitestConstants,
-				docsEnabled: (server?.ctx.custom?.toolsets?.docs ?? true) && manifestStatus.available,
+				testEnabled: (server?.ctx.custom?.toolsets?.test ?? true) && testSupported,
+				docsEnabled: (server?.ctx.custom?.toolsets?.docs ?? true) && docsAvailable,
 				changeDetectionEnabled,
 				dependencyGraphAvailable: dependencyGraphSupported,
-				reviewEnabled: reviewStatus.available,
+				reviewEnabled: reviewAvailable,
```
ghengeveld: Naming is a bit inconsistent here. Is that intentional or should we align these?

ghengeveld (approved):

packages/addon-mcp/src/mcp-handler.ts
yannbf: It's now aligned!

PR #273, opened 2026-06-04
Migrate dep graph logic to open service architecture

yannbf:
## What this does

This PR introduces changes to align with recent updates in Storybook [#35048](https://github.com/storybookjs/storybook/pull/35048) where the old
dependency-graph API that `get-stories-by-component` relied on got removed.

## What changed

- **New `utils/module-graph.ts`** — wraps `getService('core/module-graph')`, replacing the
  deleted `utils/change-detection.ts`. It treats a missing `getService` export or an
  unregistered service as "unsupported", so older Storybook versions degrade gracefully.
- **`resolve-component-stories.ts`, `detect-unreachable-changes.ts`, `get-tool-availability.ts`**
  now consume the service: check `getStatus` for readiness, then do a single batched
  `getStoriesForFiles` lookup instead of one `lookup()` per target.
- **Path handling** — the service returns story files as story-index-relative paths
  (`./src/Button.stories.tsx`), so we map story IDs off those rather than absolute paths.
  Barrel expansion, depth merging, and sorting are unchanged.
- **Tool gating** — `get-stories-by-component` is now gated on the service being *registered*
  (`isModuleGraphSupported`), not just on an export existing. On Storybook versions without
  the service, the tool is hidden and the existing "unavailable" message is shown.
- **Storybook bumped** `10.5.0-alpha.4` → `10.5.0-alpha.6` across the catalog and lockfiles.
- Tests updated to match the new path format and gating.

## Testing

I tested the canary against real projects (e.g. MealDrop) and the detection works as intended.


packages/addon-mcp/src/utils/module-graph.ts
```diff
@@ -0,0 +1,90 @@
...
+	depth: number;
+}
+
+/** A runtime open-service query: callable synchronously, or awaitable in full via `loaded`. */
+interface RuntimeQuery<TInput, TOutput> {
+	(input: TInput): TOutput;
+	loaded(input: TInput): Promise<TOutput>;
+}
+
+/** The subset of the `core/module-graph` runtime service surface that addon-mcp consumes. */
```
JReinhold, line 41: most of these types should have been importable from core Storybook. maybe they aren't, but we should fix that down the line.

JReinhold (approved):

packages/addon-mcp/src/utils/module-graph.ts
yannbf: Thanks! Now that alpha 6 was released and contains _some_ types I added `Query` and deleted one of the locally defined types. The other types (ModuleGraphErrorLike, ModuleGraphStatus, ModuleGraphStoryHit, getStoriesForFiles) are not exposed though!

ghengeveld (approved):

PR #302, opened 2026-06-19
fix(ade): skip sb installation if no explicit skill invocation

huang-julien:
This pull request updates the prerequisites for Storybook-related skills to clarify the behavior when Storybook is not installed. Now, the skills will silently ignore the absence of Storybook unless the user explicitly invokes the skill, instead of stopping and requiring user opt-in.

**Storybook Skill Prerequisite Updates:**

* Updated the prerequisite instructions in `SKILL.md` files to specify that if Storybook is not installed and the skill is not explicitly invoked by the user, the skill should be silently ignored. [[1]](diffhunk://#diff-82c7b7ac13c662c4bed1eb4a272a699530970104c45c2ef7b031363e5e39cba0L37-R37) [[2]](diffhunk://#diff-ff59803e6fadff805e759bd4b0f599ebd36e8171b385c3beac9e6f462dfc8fedL26-R26)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when Storybook isn’t installed: story-related tasks are now skipped automatically unless the skill was explicitly invoked. When invoked, the user is prompted to set up Storybook and proceed; declining will skip story work while still allowing any non-story component changes to complete.

### DIFF
--- a/packages/claude-plugin/skills/stories/SKILL.md
+++ b/packages/claude-plugin/skills/stories/SKILL.md
@@ -38,7 +38,7 @@ stories in sync with every frontend component you create, modify, or delete.
   - If this skill is not explicitly invoked by the user, silently ignore this skill.
   - Otherwise, ask the user whether they want to set up Storybook and write stories.
   - If the user decline, silently skip all story work, finish the component change (if any), and move on.
-  - If the user accepts, invoke the **storybook-setup** skill to set up Storybook and write stories.
+  - If the user accepts, invoke the **storybook-init** skill to set up Storybook and write stories.
 - Storybook must be a canary version (0.0.0-canary) or at least version 10.5. If an older version is
   installed, invoke the **storybook-upgrade** skill to upgrade it before
   proceeding.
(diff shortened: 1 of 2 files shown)


packages/claude-plugin/skills/stories/SKILL.md
```diff
@@ -34,52 +34,15 @@ stories in sync with every frontend component you create, modify, or delete.
 
 ## Prerequisites
 
-- Storybook must be installed in the project. If it is not, go to
-  "When Storybook is not installed" below and stop here unless the user opts in.
+- Storybook must be installed in the project. If it is not installed:
+  - If this skill is not explicitly invoked by the user, silently ignore this skill.
+  - Otherwise, ask the user whether they want to set up Storybook and write stories.
+  - If the user decline, silently skip all story work, finish the component change (if any), and move on.
+  - If the user accepts, invoke the **storybook-setup** skill to set up Storybook and write stories.
```
kasperpeulen: The storybook-init skill right?

kasperpeulen (approved):

packages/claude-plugin/skills/stories/SKILL.md
huang-julien: oops indeed

PR #368, opened 2026-07-22
fix(mcp): surface component @deprecated JSDoc tags in get-documentation

### LINKED ISSUE #367: get-documentation omits component JSDoc tags (@deprecated) from its output
## Summary

`get-documentation` omits component-level JSDoc tags — notably `@deprecated` — from the markdown it returns. The tag is extracted and present in the manifest/docgen data, but it is dropped when that data is formatted into the tool response. So an agent calling `get-documentation` cannot see that a component is deprecated unless the deprecation is also written into the prose description by hand.

This reproduces **both with and without** `experimentalDocgenServer`. I noticed it while testing the docgen server, but it isn't specific to that flag — both paths go through the same formatter.

**Reproduction:** https://github.com/rachelslurs/storybook-mcp-jsdoctags-repro

## Versions

- `storybook` 10.5.3
- `@storybook/react-vite` 10.5.3
- `@storybook/addon-mcp` 0.7.0 (bundles `@storybook/mcp` 0.8.0)
- Framework: `react-vite`

## Steps

A component with a neutral description and a distinctive `@deprecated` reason:

```tsx
/**
 * A basic button.
 *
 * @deprecated Use `NewButton` from `@acme/ui` instead.
 */
export const Button = ({ label, size = 'medium' }: ButtonProps) => { /* ... */ };
```

```bash
npm install
npm run storybook
node probe.mjs example-button   # does the MCP handshake and prints get-documentation's output
```

## Expected

The returned documentation surfaces that `Button` is deprecated (ideally with the reason), so an MCP consumer sees it.

## Actual

The deprecation is nowhere in the output — no `Deprecated`, no `@acme/ui`, no `NewButton`:

````md
# Button

ID: example-button

A basic button.

## Stories

### Primary

Story ID: example-button--primary

```
import { Button } from "sb-mcp-jsdoctags-repro";

const Primary = () => <Button label="Click me" />;
```

## Props

```
export type Props = {
  /**
    Button label text.
  */
  label: string;
  /**
    Visual size of the button.
  */
  size?: "small" | "medium" | "large" = medium;
}
```
````

## The data is present; it's dropped at formatting

**Flag off** — `/manifests/components.json` (dev, v0 inline):

```jsonc
"example-button": {
  "jsDocTags": {},
  "reactDocgenTypescript": { "tags": { "deprecated": "Use `NewButton` from `@acme/ui` instead." } } // string
}
```

**Flag on** — built `services/core/docgen/example-button.json`:

```jsonc
"example-button": {
  "jsDocTags": { "deprecated": ["Use `NewButton` from `@acme/ui` instead."] } // array
}
```

## Likely root cause (in `@storybook/mcp`)

- `get-documentation` returns `formatComponentManifest(component)` (`src/tools/get-documentation.ts`).
- `formatComponentManifest` emits title / `ID:` / description / subcomponents / stories / props / docs — it never reads `jsDocTags`.
- `getParsedDocgen` → `parseComponentDocLike` (shared by `reactDocgenTypescript` and `reactComponentMeta`) returns only `{ props }`, so the component-level `tags`/`jsDocTags` are dropped there too.

There's no path from the extracted tag to the formatted output in either mode. `jsDocTags` *is* carried into the internal manifest shape (`adaptCoreComponent` only drops `argTypes`), which suggests it's meant to be usable downstream — so this reads like a formatter gap rather than an intentional omission, but happy to be told otherwise.

## Suggestion

Render component `jsDocTags` (at minimum `@deprecated`, with its reason) in `formatComponentManifest`, e.g. a short `> Deprecated: <reason>` line under the component heading. Would make deprecation travel as a tag rather than requiring it be duplicated into the description.

rachelslurs:
Fixes #367. Reproduction: https://github.com/rachelslurs/storybook-mcp-jsdoctags-repro

## Summary

`get-documentation` now renders a component's `@deprecated` JSDoc tag. The tag was extracted into the manifest but never printed, so an agent calling `get-documentation` could not see that a component was deprecated unless the deprecation was also written into the prose description by hand.

## Problem

`formatComponentManifest` prints the component name, id, description, subcomponents, stories, props, and attached docs. It never reads `jsDocTags`. The tag reaches the formatter by two paths, and both drop it:

- With `experimentalDocgenServer`, the deprecation sits on the manifest's top-level `jsDocTags.deprecated` (a `string[]`). `adaptCoreComponent` passes it through, and the formatter ignores it.
- Without the flag, react-docgen-typescript puts it on `reactDocgenTypescript.tags.deprecated` (a string), and `parseComponentDocLike` reads only `props`, so it is gone before the formatter runs.

I reproduced both on a react-vite project with Storybook 10.5.3 and addon-mcp 0.7.0. The `get-documentation` output carried the description and no deprecation, flag on and flag off.

## Changes

`packages/mcp/src/utils/parse-react-docgen.ts`

- Add an optional `tags: Record<string, string[]>` to `ParsedDocgen`, populated only when the engine reports tags, so parser output is unchanged for components without tags.
- `parseComponentDocLike` (react-docgen-typescript and `reactComponentMeta`) normalizes the engine's `tags` into that field.

`packages/mcp/src/utils/manifest-formatter/markdown.ts`

- Add `formatDeprecationNotice`, which reads `jsDocTags.deprecated` first and falls back to `parsedDocgen.tags.deprecated`.
- `formatComponentManifest` and `formatSubcomponentsSection` print a `> **Deprecated:** <reason>` line under the component or subcomponent heading, above the description. A tag with no message prints a bare `> **Deprecated**`. A component with no tag, or an empty `deprecated` array, prints nothing. Non-deprecated components render byte for byte as before.

Output for a deprecated `Button`:

````md
# Button

ID: example-button

> **Deprecated:** Use `NewButton` from `@acme/ui` instead.

A basic button.
````

## Scope

- Covers the docgen-server (`jsDocTags`) and react-docgen-typescript / `reactComponentMeta` (`tags`) paths, for components and subcomponents, verified end to end.
- react-docgen (the JS/Flow engine) is not covered. Its base `Documentation` type carries no component-level `tags`, and this code does not read tags if Storybook attaches them downstream.
- `list-all-documentation` still shows only the component summary line. Deprecation appears once you fetch the component with `get-documentation`.

## Test plan

- [x] `pnpm vitest --project=@storybook/mcp`: 205 pass, including new tests for both resolution paths, the bare-notice case, the empty-array guard, subcomponent deprecation, and a guard that a non-deprecated component renders unchanged.
- [x] `pnpm --filter @storybook/mcp typecheck` and `oxlint --type-aware`: clean.
- [x] Built `@storybook/mcp`, loaded it into the reproduction project, and confirmed the callout appears on the flag-on and flag-off paths.

## Summary by CodeRabbit

* **Documentation**
  * Component and subcomponent documentation now renders JSDoc `@deprecated` notices as callouts beneath the relevant headings.
  * Deprecation reasons are formatted consistently and supported from both current and legacy metadata shapes.
  * Empty deprecation messages are shown without a reason; missing/empty deprecation values correctly omit the notice.
  * Subcomponent deprecation notices are displayed in-place directly under each subcomponent heading.
* **Tests**
  * Added coverage for deprecation rendering and tag normalization behavior.

### DIFF
--- a/.changeset/deprecated-jsdoc-tags-in-get-documentation.md
+++ b/.changeset/deprecated-jsdoc-tags-in-get-documentation.md
@@ -1,0 +1,11 @@
+---
+"@storybook/mcp": patch
+---
+
+Surface component `@deprecated` JSDoc tags in `get-documentation` output
+
+Component-level JSDoc tags were extracted into the manifest but never rendered by the
+markdown formatter, so `get-documentation` silently dropped `@deprecated`. It is now
+shown as a `> **Deprecated:** <reason>` callout under the component and subcomponent
+headings, resolved from the docgen-server manifest (top-level `jsDocTags.deprecated`) and
+from react-docgen-typescript / `reactComponentMeta` output (the engine's `tags.deprecated`).
--- a/packages/mcp/src/utils/manifest-formatter/markdown.test.ts
+++ b/packages/mcp/src/utils/manifest-formatter/markdown.test.ts
@@ -70,6 +70,113 @@ describe('MarkdownFormatter - formatComponentManifest', () => {
 		});
 	});
 
+	describe('deprecation notice', () => {
+		it('renders @deprecated from top-level jsDocTags (docgen-server path), above the description', () => {
+			const manifest: ComponentManifest = {
+				id: 'button',
+				name: 'Button',
+				path: 'src/components/Button.tsx',
+				description: 'A basic button.',
+				jsDocTags: { deprecated: ['Use `NewButton` from `@acme/ui` instead.'] },
+			};
+
+			const result = formatComponentManifest(manifest);
+
+			expect(result).toContain('> **Deprecated:** Use `NewButton` from `@acme/ui` instead.');
+			expect(result.indexOf('> **Deprecated:**')).toBeLessThan(result.indexOf('A basic button.'));
+			expect(result).toMatchInlineSnapshot(`
+				"# Button
+
+				ID: button
+
+				> **Deprecated:** Use \`NewButton\` from \`@acme/ui\` instead.
+
+				A basic button."
+			`);
+		});
+
+		it('renders @deprecated from react-docgen-typescript tags (legacy path)', () => {
+			const manifest: ComponentManifest = {
+				id: 'button',
+				name: 'Button',
+				path: 'src/components/Button.tsx',
+				description: 'A basic button.',
+				reactDocgenTypescript: {
+					props: {},
+					tags: { deprecated: 'Use `NewButton` from `@acme/ui` instead.' },
+				},
+			};
+
+			const result = formatComponentManifest(manifest);
+
+			expect(result).toContain('> **Deprecated:** Use `NewButton` from `@acme/ui` instead.');
+		});
+
+		it('renders a bare notice when the tag has no message', () => {
+			const manifest: ComponentManifest = {
+				id: 'button',
+				name: 'Button',
+				path: 'src/components/Button.tsx',
+				jsDocTags: { deprecated: [''] },
+			};
+
+			const result = formatComponentManifest(manifest);
+
+			expect(result).toContain('> **Deprecated**');
+			expect(result).not.toContain('> **Deprecated:**');
+		});
+
+		it('omits the notice entirely for non-deprecated components', () => {
+			const manifest: ComponentManifest = {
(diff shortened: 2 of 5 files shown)


valentinpalkovic:
Superseded by https://github.com/storybookjs/storybook/pull/35963

PR #374, opened 2026-07-28
Agentic reference eval base work

yannbf:
#### What
This PR introduces a new eval line (7xx) for research runs: an agent does a small task in a real external repo, with a published design-system Storybook MCP available, and we measure whether the MCP changed what it did.

`external-repo.ts` — Makes it possible to download a GitHub repo into the sandbox at a pinned SHA. You just declare it in the package.json e.g. `evals.externalRepo: { repo, ref }`. It takes care of installing the dependencies on whichever package manager, and making sure that the vitest config file doesn't conflict with the user's current vitest config file.

`external-mcp.ts` — Registers an externally hosted Storybook MCPs (like from Chromatic) in Claude Code's `.mcp.json` or Codex's `.codex/config.toml`, under the same server name the stock templates use, so the existing #test-utils workflow helpers work as well. If the MCP URL is dead, it fails before any agent tokens are spent.

`/lib/agentic-reference/experiment.ts` — Where the shape of the agentic reference experiment lives.

#### How to use

To execute the evals:

Claude code:
```EVAL_AGENTIC_REFERENCE=1 pnpm exec agent-eval agentic-ref-reuse-component-mcp-cc-opus-high```

Codex:
```EVAL_AGENTIC_REFERENCE=1 pnpm exec agent-eval agentic-ref-reuse-component-codex-mcp-gpt-5.5-medium```

If you need to run the same eval N times, you can use `AGENTIC_REF_RUNS` e.g.
```EVAL_AGENTIC_REFERENCE=1 AGENTIC_REF_RUNS=10 pnpm exec agent-eval agentic-ref-reuse-component-cc-mcp-opus-high```

👆 That will run the eval **test** and produce the **standard** result output from Vercel. To get the extra data, e.g. design system coverage etc that we are planning on doing, run `pnpm results:analyze` which will do that job. It currently is barebones and does one metric: Button imports before/after/delta. Everything else is a stub.

## Summary by CodeRabbit

- **New Features**
  - Added an agent evaluation scenario focused on component reuse and confirming design-system documentation tooling usage.
  - Introduced offline analysis of agent-evaluation results, generating per-run and aggregated metrics/cost summaries.
  - Added support for materializing and evaluating external repository snapshots with pinned refs.
- **Enhancements**
  - Extended pricing for the latest Claude model to improve cost estimation.
  - Added automated experiment configuration for agentic-reference runs and MCP tool-usage reporting.
- **Chores**
  - Prevented cached evaluation data and result directories from being collected as tests.
  - Updated ignore rules to skip eval cache output.

### CODE REVIEW

File: agent-eval/experiments/agentic-ref-reuse-component-codex-mcp-gpt-5.5-medium.ts
Comment (Sidnioulz) [inline agent-eval/experiments/agentic-ref-reuse-component-codex-mcp-gpt-5.5-medium.ts:1]: nit: I'd suggest removing until we use it for real. One less file to maintain in the meantime.

File: agent-eval/evals/701-agentic-ref-reuse-component-mcp/EVAL.ts [hunk 1]
```diff
@@ -0,0 +1,81 @@
+// Agentic-reference reuse-component eval: the agent makes a small footer change
+// that should reuse the app's own Button component, with the design system's
+// published Storybook MCP available.
+// Asserts the outcome and one MCP-usage signal (published `@storybook/mcp`
+// builds expose only the documentation workflow, hence
+// expectDocumentationToolingCalled). Heavy, dep-needing metrics (the app's own
+// test suite, baseline vs after) run offline in scripts/analyze-results.mjs.
```
Sidnioulz: Tempted to just delete all these comments. The workflow is due to change a little so they'll instantly expire.

agent-eval/lib/agentic-reference/experiment.ts
```diff
@@ -0,0 +1,107 @@
...
+const AGENT_CONFIG: Record<EvalAgent, AgentConfig> = {
+	'claude-code': {
+		agent: 'vercel-ai-gateway/claude-code',
+		model: 'opus',
+		agentOptions: { effort: 'high' },
+	},
+	codex: {
+		agent: 'codex',
+		model: 'gpt-5.5?reasoningEffort=medium',
+	},
```
Sidnioulz, line 90:
```suggestion
	// codex: {
	// 	agent: 'codex',
	// 	model: 'gpt-5.5?reasoningEffort=medium',
	// },
```

agent-eval/lib/agentic-reference/experiment.ts
```diff
@@ -0,0 +1,107 @@
...
+};
+
+/** Research sample size, from AGENTIC_REF_RUNS (default 1). */
+function resolveRuns(): number {
+	const raw = process.env.AGENTIC_REF_RUNS;
+	if (raw === undefined || raw === '') {
+		return 1;
```
Sidnioulz:
```suggestion
/** Research sample size, from AGENTIC_REF_RUNS (default 10). */
function resolveRuns(): number {
	const raw = process.env.AGENTIC_REF_RUNS;
	if (raw === undefined || raw === '') {
		return 10;
```

agent-eval/lib/agentic-reference/external-mcp.ts
```diff
@@ -0,0 +1,113 @@
...
+		throw new Error(`registerExternalStorybookMcp: ${url} is unreachable: ${String(error)}`);
+	}
+	if (!response.ok) {
+		throw new Error(
+			`registerExternalStorybookMcp: ${url} answered HTTP ${response.status} to an MCP initialize request`,
+		);
+	}
+}
+
+async function registerForClaude(sandbox: Sandbox, url: string): Promise<void> {
```
Sidnioulz: Could be converted to a generic `registerExternalMcp` util.

agent-eval/lib/agentic-reference/external-repo.ts
```diff
@@ -0,0 +1,141 @@
...
+
+	await installDependencies(sandbox);
+
+	const appVitestConfig = await sandbox.readFile('vitest.config.ts').catch(() => null);
+
+	const appManifest: unknown = JSON.parse(await sandbox.readFile('package.json'));
+	if (!isRecord(appManifest)) {
+		throw new Error(`setupExternalRepo: ${repo}@${ref} package.json is not a JSON object`);
+	}
+	appManifest.imports = {
```
Sidnioulz, line 239:
```suggestion
  // Ensures EVAL.ts subpath imports are defined in the fetched repo.
	appManifest.imports = {
```

agent-eval/lib/agentic-reference/external-repo.ts
```diff
@@ -0,0 +1,141 @@
...
+		...(isRecord(appManifest.imports) ? appManifest.imports : {}),
+		'#test-utils': './__agent_eval__/test-utils.ts',
+	};
+	appManifest.devDependencies = {
```
Sidnioulz, line 244:
```suggestion
  // Adds the agent eval dependency to the repo.
	appManifest.devDependencies = {
```

agent-eval/lib/agentic-reference/external-repo.ts
```diff
@@ -0,0 +1,141 @@
...
+		...(isRecord(appManifest.devDependencies) ? appManifest.devDependencies : {}),
+		'@vercel/agent-eval': harnessVersion(),
+	};
+
+	const existingNpmrc = await sandbox.readFile('.npmrc').catch(() => '');
```
Sidnioulz, line 252:
```suggestion
  // In some external repos (e.g. Mealdrop), there might be peer dependency
  // issues that we don't care about and that can cause the eval to fail
  // installing deps. Silently ignore them.
	const existingNpmrc = await sandbox.readFile('.npmrc').catch(() => '');
```

Sidnioulz (approved):

agent-eval/lib/agentic-reference/experiment.ts
yannbf: This is just a schema definition, there's no harm in leaving that in

agent-eval/lib/agentic-reference/experiment.ts
yannbf: Not doing that yet, but I added a TODO in code comments

PR #377, opened 2026-07-29
Add agentic-reference initial metrics

Sidnioulz:
Per-run quality metrics for the agentic-reference evals, plus two harness fixes.

## Metrics

| metric | source |
| --- | --- |
| duration, turns | `result.json` |
| input/output/cache tokens, cost, cache-hit rate | `metadata.usage` |
| tool-call counts, five-bucket taxonomy ( `docs` / `exploration` / `edit` / `verification` / `other`) | `transcript.json` |
| per-file edit iterations, including shell writes and renames | `transcript.json` |
| SLoC diff, comment- and blank-stripped | pinned ref vs collected tree |
| cyclomatic + cognitive complexity diff | TypeScript compiler API, per-sha baseline |

Written to `analysis.json` beside each `result.json`, aggregated into `results/agentic-ref-analysis.json`.

All metrics are computed from stored artifacts only, so the pass re-runs over historical results without a sandbox or model call. A metric that can only be computed during a run fixes its own definition at run time.

## Architecture

`scripts/analyze-results.mjs` discovers runs and dispatches to a per-eval hook at `evals/<name>/post-analysis.ts`. It contains no eval-specific logic. Evals without a hook are skipped (47 of 48).

```
evals/701-agentic-ref-reuse-component-mcp/
  post-analysis.ts     analyzeRun / summarize / renderTables
  __analysis__/        metric modules, tests, fixtures, baselines
```

Adding metrics to another eval means adding a `post-analysis.ts` there.

TypeScript modules are imported from the `.mjs` gateway via Node 24 type stripping, so metric code is typed and tested while the script keeps no build step.

## Harness fixes

Both in `patches/@vercel__agent-eval@1.2.0.patch`.

1. Binary corruption on collection. `captureGeneratedFiles` read files with `sandbox.readFile`, which returns stdout as a string; the sandbox decodes that stream as UTF-8 per chunk. `favicon.ico` came out 14,254 → 24,506 bytes with 5,125 replacement characters. All images and fonts in collected trees were affected. Now reads via `base64` on both backends. A second lossy path in `readFixtureFiles` is fixed too.

2. Analysis code reached the agent's workspace. Fixture directories are uploaded to the sandbox, so `post-analysis.ts` and `__analysis__/` were readable by the agent under evaluation. Added to `IGNORED_PATTERNS` and `EXCLUDED_FILES`.

## Review

The diff is +17,931 lines; most is a transcript fixture, a complexity baseline, docs, and an archived prototype. Implementation is ~3,000 lines and tests ~1,300, in `evals/701-*/__analysis__/`.

Reading order: `paths.ts`, `shell-segments.ts`, `tool-taxonomy.ts`, `tree-diff.ts`, `post-analysis.ts`, `scripts/analyze-results.mjs`.

No eval run:

```bash
pnpm --filter agent-eval typecheck
npx vitest run --project agent-eval        # 207 tests, ~150 new
pnpm lint                                  # 14 warnings / 0 errors is the baseline
pnpm format:check
```

Against a stored run (fetches the pinned ref, ~20 MB, cached):

```bash
cd agent-eval && pnpm results:analyze

cp results/agentic-ref-analysis.json /tmp/p1.json
pnpm results:analyze >/dev/null && diff /tmp/p1.json results/agentic-ref-analysis.json
```

The diff should be empty.

The corruption fix is verified offline against a fake `@vercel/sandbox` reproducing the chunked stdout; scripts are in commit `fafeedf`.

With an eval run (~$2, ~7 min):

```bash
cd agent-eval
EVAL_AGENTIC_REFERENCE=1 AGENTIC_REF_RUNS=1 pnpm eval agentic-ref-reuse-component-cc-mcp-opus-high

# should return nothing
find results -path "*/run-*/project/post-analysis.ts" -o -path "*/run-*/project/__analysis__*"

# should equal the ref's 14254
stat -c%s results/**/run-1/project/public/favicon.ico
```

If the `find` returns anything, metrics from that run are invalid: the agent could read what it was scored on.

## Things to check

- The patch touches 14 files in the vendored `dist/` and needs re-verifying on every `@vercel/agent-eval` upgrade. A failing hunk shows at `pnpm install`; a hunk applying cleanly to changed upstream code does not. The corruption fix is being upstreamed and can be dropped from the patch when it lands.
- `generatedFiles` changed from `Record<string, string>` to `Record<string, Buffer>`. Hooks reading it must decode.
- Exclusion lists in `paths.ts` and `IGNORED_PATTERNS` match on basename. Generic entries would drop application source; `helpers` was rejected because the app ships `src/helpers/`.
- Metrics that cannot be computed are `null`, not `0`. Ratios with a zero denominator are `null`, not `Infinity`.
- Baselines at `__analysis__/baselines/<repo>@<sha>.json` are keyed by pin, so a pin bump adds a file rather than reusing stale numbers. Vendored files are excluded: `.yarn/releases/yarn-4.2.1.cjs` was 98% of the repo's cyclomatic total.
- The CI step is `continue-on-error: true`. A broken analyzer produces no `analysis.json` and no red build.
- Cyclomatic charges +1 per function for existing, so extracting helpers reads as worse; cognitive does not. Prefer cognitive when they disagree. Per-function mean is not implemented — SonarQube removed that metric in 2025.4.

## Limitations

- The corruption fix has not run against a live Vercel sandbox. Verification is offline. `base64` must exist in the sandbox image, and capture sends ~33% more bytes.
- Trees collected before the fix cannot be repaired. The corruption is chunk-dependent (24,506 bytes in one run, 24,500 in another), so it is not invertible.
- Setup replaces the app's berry `yarn.lock` with a v1 file containing harness dependencies, and generates a `package-lock.json` with 22 more. Not agent-caused: neither captured run ran an install. No metric here is affected, but a run where the agent adds a dependency is invisible, and the collected tree cannot be built with its own package manager. Tracked separately.
- Accessibility metrics are out of scope. A Playwright + axe-core prototype and its measurements are archived at `docs/superpowers/prototypes/2026-07-29-a11y-journeys/`.
- Churn counts scratch files, so it does not reconcile with the SLoC diff, which only sees files that survived.


agent-eval/baselines/701-agentic-ref-reuse-component-mcp/yannbf__mealdrop@ce507b345666ea8678101fccac580186b2b69b1f.json
Sidnioulz, line 1: This is the precomputed metrics data for the baseline external repo. It's cached automatically and committed to save compute costs.

agent-eval/evals/701-agentic-ref-reuse-component-mcp/EVAL.ts
Sidnioulz, line 1: I tidied this file up as there were lots of noise inside.

agent-eval/lib/agentic-reference/__fixtures__/golden-run/result.json
Sidnioulz, line 1: This file and the transcript are an old run captured previously that acts as a test fixture. It might drift in the future though.

agent-eval/lib/agentic-reference/external-repo.ts
```diff
@@ -39,6 +48,60 @@ export function parseExternalRepoMarker(packageJsonContent: string): ExternalRep
...
+
+const refCache = new Map<string, string>();
+
+/**
+ * Download and extract a ref, returning its directory. Extraction happens in a
+ * scratch directory that is renamed into place only once it fully succeeded: a
+ * half-populated cache directory would be trusted forever and would quietly
+ * skew every later diff.
+ */
+export function prepareRef(cacheDir: string, repo: string, ref: string): string {
```
Sidnioulz, line 78: Useful to compute the baseline run

agent-eval/patches/@vercel__agent-eval@1.2.0.patch
Sidnioulz, line 1: Needed to address the PNG corruption bug until upstream merges my fix PR.

agent-eval/scripts/analyze-results.ts
```diff
@@ -0,0 +1,384 @@
...
+// each run to the module its *experiment* names in `postAnalysis` — see
+// lib/post-analysis/types.ts. Runs whose experiment names none are skipped.
+//
+// Keying on the experiment rather than the eval is what lets a family of arms —
+// same task and setup, differing only in prompt or MCP endpoint — share one set
+// of metrics, and lets one experiment span several evals.
+//
+// summarize is scoped to one eval directory — the folder holding that eval's
+// run-* dirs — and its rows are written back into that folder's summary.json
+// under `postAnalysis`, next to the harness's own pass rate and mean duration.
```
Sidnioulz, line 14: Kept it verbose because this will be relevant for the bit that describes all treatment cases: we still need to ensure the summary summarizes the N runs with the same conditions. The current FS architecture that guarantees this may change with the API changes we've discussed.

.github/workflows/agent-eval.yml
```diff
@@ -121,6 +121,14 @@ jobs:
...
           echo "has_result_files=true" >> "$GITHUB_OUTPUT"
 
+      - name: Compute offline metrics
+        if: ${{ always() && steps.check_results.outputs.has_result_files == 'true' }}
+        # Write analysis.json next to each run's result.json so the metrics ride
+        # along in the uploaded artifact.
+        continue-on-error: true
+        working-directory: agent-eval
+        run: pnpm results:analyze
+
```
Sidnioulz, line 131: I need to see how this actually behaves. Of course, we need the post-analysis results to be committed for them to be preserved long-term.

agent-eval/lib/agentic-reference/metrics/tool-taxonomy.ts
```diff
@@ -0,0 +1,255 @@
...
+	'pkill',
+	'wait',
+	'export',
+	'cd',
+]);
+
+/** Wrappers to step past to reach the binary that actually runs. */
+const PACKAGE_RUNNERS = new Set(['npx', 'npm', 'pnpm', 'yarn', 'bun', 'bunx']);
+// `xargs` is deliberately absent: its own arguments interleave with the command
+// it runs (`xargs -n 1 rm`), so stepping past it cannot be done by flag-skipping.
```
Sidnioulz: getopt algorithm can strip args properly if we have the signature for xargs.

agent-eval/lib/utils/shell-segments.ts
```diff
@@ -0,0 +1,120 @@
...
+	redirectTarget: string | null;
+	/** This segment consumes the previous segment's stdout. */
+	piped: boolean;
+}
+
+const SEPARATORS = new Set(['&&', '||', ';', '|']);
+
+// Heredoc bodies are data, not commands. Left in place, a payload containing
+// `rm -rf /` would be tokenised and classified as an edit.
+const HEREDOC = /<<-?\s*'?"?(\w+)'?"?[\s\S]*?^\1$/gm;
```
yannbf: The terminator regex requires column 0, but <<- allows tab-indented terminators. The heredoc body then gets tokenized as real commands — the exact failure mode the file's own header comment says the stripping exists to prevent.

Reproduced failure
computeChurn([shell('cat > file.sh <<-EOF\n\trm -rf /workspace/important\n\tEOF')]).perFile
// => { "file.sh": 1, "important": 1 }   — heredoc *data* counted as an edit

agent-eval/lib/agentic-reference/post-analysis.ts
```diff
@@ -0,0 +1,257 @@
...
+			// Complexity correlates ~0.9 with lines of code, so a bare delta partly
+			// re-measures verbosity. null rather than Infinity when nothing changed:
+			// a stored Infinity would poison every later mean.
+			densityPerSloc: diff.sloc.net === 0 ? null : cognitiveDelta / diff.sloc.net,
+			parseFailures: afterFiles.parseFailures,
+		},
+	};
+}
+
+function mean(values: number[]): number | null {
```
Sidnioulz: This should be shared utils.

In particular, the math package does exist.

agent-eval/lib/agentic-reference/post-analysis.ts
```diff
@@ -0,0 +1,257 @@
...
+export function deltaToBaseline({
+	baselineAnalysis,
+	baselineDir,
+	// TODO: add projectScore
+	projectDir,
+}: DeltaToBaselineContext): Analysis {
+	const diff = diffTrees(baselineDir, projectDir);
+
+	// Complexity needs an AST, so .css participates in SLoC but not here.
+	const changedScripts = diff.files.filter((file) => SCRIPT_EXTENSIONS.test(file));
```
Sidnioulz: Might be useless, the complexity metric should handle it.

agent-eval/lib/agentic-reference/post-analysis.ts
```diff
@@ -0,0 +1,257 @@
...
+
+	// Both sides are scored per file, then totalled, so they stay comparable.
+	const afterFiles = complexityForFiles(projectDir, changedScripts);
+	const before = sumComplexities(scoresFor(baselineFiles(baselineAnalysis), changedScripts));
+	const after = sumComplexities(afterFiles.files);
```
Sidnioulz: This is wrong, the `after` sum needs to be all the baseline complexities - the baseline changed files' complexities + the treatment changed files' complexities.

agent-eval/evals/701-agentic-ref-reuse-component-mcp/EVAL.ts
```diff
@@ -1,74 +1,7 @@
...
-	'get-documentation',
-	'get-documentation-for-story',
-	'list-all-documentation',
-];
+import { getTranscript } from '#test-utils';
 
 test('agent produced a transcript', () => {
 	const transcript = getTranscript();
 	expect(transcript.events.length, 'Expected the transcript to contain events').toBeGreaterThan(0);
 });
```
Sidnioulz, line 7: We wanna keep this _and_ make test failures actual errors so we can see in Vercel when a run failed and must be redone.

yannbf (approved):

PR #394, opened 2026-08-14
Sync the agentic-ref eval workflow from the eval branch

yannbf:
The registered copy of `agentic-ref-eval.yml` on main had drifted behind the version the eval branch actually runs (dispatches execute the selected ref's file, but the two should not disagree). One file, syncing:

- **Credential validation before spending** — a missing `AI_GATEWAY_API_KEY` makes the runner skip every eval with exit 0, so the job went green having run nothing; the gate now fails up front.
- **`expect` spend guard** — state how many evals the selection should resolve to; a free dry-run checks the real plan first and the job fails before any spend on mismatch. Verified live: a deliberate `expect=3` against a 15-eval plan fails with both numbers ([run](https://github.com/storybookjs/mcp/actions/runs/31806873893)); a matching expectation passes ([run](https://github.com/storybookjs/mcp/actions/runs/31811201306)).
- **Flow selection by number or glob** — `flows=703` resolves to `703-fix-bug-flow`, `flows=70*` to every *active* flow; tokens expand against `AGENTIC_REF_EVALS`, not the evals directory, so parked fixtures can't sneak into a selection, and an unknown token fails the resolve step with the active list in the message. Verified live for both forms.
- **Comma-separated `cases`**, matching `flows`.
- **`runs` defaults to 10**, the batch size the comparison tooling expects.
- Plain-language input descriptions.

Dispatching still requires `--ref agentic-reference-eval` — main carries the registration, the eval scripts live on the branch (as the file header documents).

### DIFF
--- a/.github/workflows/agentic-ref-eval.yml
+++ b/.github/workflows/agentic-ref-eval.yml
@@ -14,25 +14,29 @@ on:
   workflow_dispatch:
     inputs:
       cases:
-        description: 'Space-separated case names or globs, e.g. agentic-ref-cc-control-none-opus-high agentic-ref-cc-docs-full-opus-high — empty runs every case'
+        description: 'Comma-separated case names or globs, e.g. agentic-ref-cc-control-none-opus-high,agentic-ref-cc-docs-full-opus-high — empty runs every case'
         type: string
         default: ''
       flows:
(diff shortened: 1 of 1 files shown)


Sidnioulz (approved):
There'd be value in centralising all the case / flow / experiment selection in nodejs rather than bash, so the CLI uses it. Let's see about that next week if we feel it's useful for our day to day exploration of results.

LGTM

PR #398, opened 2026-08-17
Add the DS misuse metric: an LLM judge for design-system decisions

Sidnioulz:
## What this adds

`ds-coverage` measures **how much** of a run's UI comes from the design system. It cannot say whether the agent used it **well**. Two arms can land on identical DS share while one made good decisions and the other wrapped `<Card>` around markup that wanted `<PageSection>`.

`ds-misuse` measures the decisions. It is an LLM judge over the JSX nodes a run introduced, scored against the Droppy design system's own documentation. Three sub-scores, each `1` / `0.5` / `0` per node, summarised as a mean in `[0, 1]`:

| Score | Asked of | Question |
|---|---|---|
| `correct-ds-decision` | new DS usages | Right component, or did a better DS alternative exist? |
| `correct-ds-usage` | new DS usages | Does this violate a documented guideline? |
| `correct-local-decision` | new local usages | Should this have been local, or did a DS component with a relevant API exist? |

```bash
pnpm results:analyze --recompute   # builds baselines + the node census
pnpm format                        # sidecar JSON needs it
pnpm judge:ds-misuse --latest      # ~$0.10-0.15 per run after the first
pnpm results:analyze --misuse      # reads the artifacts, prints the tables
```

## Three decisions worth a reviewer's attention

**Every arm is judged against one pinned, complete copy of the DS docs** — deliberately *not* the docs variant that arm was served. Content variation between arms is the round's independent variable; judging each arm against what it saw would score a degraded arm against a lowered bar.

**The model decides new-versus-moved**, not code. It receives the baseline node list, the treatment node list and the diff, and buckets them itself. A deterministic multiset difference over AST paths was considered and rejected. The consequence is accepted and recorded: bucketing is not reproducible across two invocations, so the artifact stores the buckets the judge chose, not just the scores.

**`KNOWN-ISSUES.md` is not an input.** The ticket specified a known-violations exclusion list, but that catalog was written against a Mealdrop built on a different DS package than the fixtures now pin. It is stale, so every violation counts.

## Structural changes beyond the new metric

- **Baselines are re-keyed on the external-repo pin** rather than eval-plus-pin. `baselines/701-new-ui-flow/…json` and `baselines/703-fix-bug-flow/…json` were byte-identical apart from an `eval` field, because the agentic-reference baseline branch reads only `projectDir` and `pin`. `evalName`/`fixtureDir` are gone from `BaselineContext` so a module cannot quietly reintroduce eval-dependence. `metricsVersion` → 7.
- **`ds-coverage` gains an opt-in per-node census** (`includeNodes`, `--nodes`), addressed by an AST path carrying no line offsets so a node pushed down a file keeps its identity.
- **Run discovery is extracted** to `lib/post-analysis/discovery.ts`, shared by both CLIs so they cannot disagree about what a run is.
- **`analyze-results.ts` stays free.** It only reads what the judge wrote; it never makes a paid call. The artifact is merged **after** the analysis cache lookup — folding it in would cache a scoreless row that no later run would refresh without `--recompute`.

## Known limitations, documented in code

- The `#n` suffix disambiguating identical sibling paths is **positional**, so inserting a component above its twins renumbers them. Safe only because the judge decides new-versus-moved; `node-path.ts` carries an explicit "do not diff path sets" warning.
- The path chain follows JSX ancestors only, so JSX reached through a `.map()` callback starts a fresh chain. Deliberate — nesting through child expressions but not attribute ones is the only correct widening, and the loss is one link, not a subtree.

## Testing

606 passing, typecheck clean, lint clean, formatting clean. Every judge test mocks the SDK; **no test touches the network or the API**.

Verified beyond the unit tests: the DS pin resolves to exactly 43 documents over a real fetch; the analyse → judge → re-analyse ordering surfaces scores from a genuinely cached row; a restricted census produces paths identical to the whole-tree run; and the CLI's abort paths were each exercised against a synthetic results tree.

## Not yet exercised

**No real API call has been made.** The first live run is the first exercise of request assembly and prompt caching — check `usage.cache_read_input_tokens` on the second run to confirm the ~95k-token doc prefix is being read rather than rewritten.

Design notes: `docs/superpowers/specs/2026-08-14-ds-misuse-metric-design.md`. Implementation plan, with every deviation recorded: `docs/superpowers/plans/2026-08-14-ds-misuse-metric.md`.


agent-eval/lib/agentic-reference/metrics/ds-coverage/react/node-path.ts
```diff
@@ -0,0 +1,147 @@
...
+//   `render` rather than the class.
+// - The `#n` suffix that separates otherwise-identical paths is positional — it
+//   counts visit order within the file — so inserting a twin ahead of its
+//   siblings shifts every later suffix down, which is the very fragility that
+//   dropping offsets was meant to remove. The scheme survives that because the
+//   alternatives are worse (prop signatures cannot separate two identically
+//   propped buttons; hashing source text would make a prop edit read as a new
+//   node) and because nothing here diffs path sets in code. The three shapes
+//   above sharpen it, since each one manufactures the collisions `#n` breaks.
+//
```
Sidnioulz, line 34: As we need LLM judgement to reconcile a/b nodes, maybe a list format is not the right one, and maybe the format should be AST-like?

Sidnioulz:
## Thermo-nuclear code quality review — focus: code reuse

Reviewed `e295dea…3476494` (38 files, ~2.8k lines of code once the plan/spec docs are set aside).

First, the thing the PR description asks about: **the request assembly is correct.** I checked it against the current Messages API contract rather than the code's own comments — `claude-opus-4-8`, `thinking: {type:'adaptive'}`, `output_config.effort`, `output_config.format.json_schema`, `cache_control` with `ttl:'1h'`, `stop_reason === 'refusal'` + the `stop_details?.category` guard, and `messages.stream(...).finalMessage()` are all valid and correctly shaped on `@anthropic-ai/sdk ^0.117`. The schema respects the documented restrictions (`additionalProperties: false` on every object, no recursion, no numeric constraints). The cache-prefix ordering argument in `context.ts` is right, and the ~95k-token corpus is comfortably over Opus 4.8's 1024-token cacheable minimum. The `refusal`-before-`content[0]` ordering in `judge.ts` is the detail most callers get wrong, and it's right here. I have no API-surface findings.

The findings below are about structure.

---

### 1. Blocking — the node sidecars this metric depends on are never committed, so `judge:ds-misuse` skips every run on a fresh clone

`baselines/ds-nodes/` does not exist in the tree at this head. Both committed baselines also predate the version bump:

- `baselines/yannbf__mealdrop@ce507….json` — no `metricsVersion` field at all
- `baselines/yannbf__mealdrop@refs__tags__agentic-reference__base-ui-v1.json` — `"metricsVersion": 5`

against `postAnalysis.metricsVersion = 7`. `readNodeSidecar` requires an exact version match, so on a clean checkout it returns `null` for every pin and `judgeOne` bails at `scripts/judge-ds-misuse.ts:116-123` with *"no node census … Run: pnpm results:analyze --recompute"*. Every run, every time. The headline feature of the PR cannot run on what's committed.

`git diff --name-status -M` shows why: the two surviving baselines are `R099` renames — they were moved and had the `eval` field dropped, not regenerated. So the version bump landed in code but not in the artifacts.

That also contradicts the invariant `lib/post-analysis/baseline.ts` opens with — *"Baselines are committed rather than left in the gitignored `.eval-cache/`, so CI never recomputes one."* At 5-vs-7 the committed files are a guaranteed cache miss, so CI re-measures every pinned tree on every invocation and leaves the working tree dirty afterwards. The new `nodeSidecar` bookkeeping (`baseline.ts:230-242`) is careful about half-a-pair-on-disk, which makes it more conspicuous that the pair was never committed at all.

Fix is the workflow the PR description already documents: `pnpm results:analyze --recompute`, `pnpm format`, and commit the result — `baselines/ds-nodes/` included.

### 2. Run discovery is shared; post-analysis resolution isn't — and that's the half that decides what a run costs

`lib/post-analysis/discovery.ts` was extracted with an explicit rationale: the two CLIs *"must agree about what a run is … or the two will quietly disagree about which runs they covered."* That argument doesn't stop at discovery. `analyze-results.ts` resolves the post-analysis module **per experiment** through `lib/post-analysis/hooks.ts` (`loadPostAnalysis`, line 137), and skips runs whose experiment names none — *"not ours to measure."* `judge-ds-misuse.ts:36` instead hardcodes `import { postAnalysis } from '#lib/agentic-reference/post-analysis'` and uses that module's `metricsVersion` for every run it finds, whatever experiment produced it.

Two consequences, one of which costs money:

- The judge has no equivalent of the no-hook skip, so it will spend a model call on runs `analyze-results` deliberately declines to measure.
- `isStale` and `readNodeSidecar` are both keyed on a `metricsVersion` that is only correct because exactly one module exists today. The moment a second one does, the judge reads the wrong sidecar and reports "no node census" for a census that is sitting on disk.

`hooks.ts` says the IO half *"stays in scripts/analyze-results.ts"* — reasonable when there was one script. There are two now, and the second sidestepped the seam rather than widening it. Move `loadPostAnalysis` next to `findRuns`/`selectRuns` and let both CLIs go through it; the header comment on `discovery.ts` is already the argument for doing so.

Separately, this is a layering leak worth naming on its own: a generic `scripts/judge-ds-misuse.ts` reaching directly into `#lib/agentic-reference/post-analysis` is feature logic in a shared path.

### 3. CLI argument parsing is now duplicated three ways, and the codebase already has a canonical answer

`judge-ds-misuse.ts:54-68` is a copy of the hand-rolled `arg.split('=')` loop in `analyze-results.ts:94-125`, minus the table-section handling. Meanwhile its sibling in the same directory, `scripts/ds-coverage.ts:21,30`, uses Node's built-in `parseArgs` from `node:util`.

The reuse move is right there and unclaimed: `discovery.ts:21` already declares

```ts
export interface RunSelection { experiment: string | null; since: string | null; latest: boolean }
```

— exactly the flags both CLIs parse and feed straight into `selectRuns`. A `parseRunSelection(argv)` beside it, built on `node:util` `parseArgs`, would leave each script parsing only the flags that are actually its own (`--recompute`, the `--general/--complexity/--coverage/--misuse` family). As it stands the two parsers can drift on `--since` semantics with nothing to catch it, and both silently reject the space-separated form (`--experiment foo` reports *"Unknown argument"*).

Same file, two smaller instances of the same habit:

- `messageOf` (`judge-ds-misuse.ts:81-83`) is copied verbatim from `analyze-results.ts:127-129`, where it already existed before this PR. `lib/utils/` is right there.
- `pinOf` (`judge-ds-misuse.ts:71-79`) re-does the `isRecord` → `typecheckExternalRepo` → `catch → null` dance from `analyze-results.ts:190-197`, carrying a paraphrase of the same "the pin the run recorded, never today's" comment. The judge's version is just `pinOf(readJson(runDir + '/result.json'))` — export one and compose.

### 4. Baseline mode analyses the same tree twice

`lib/agentic-reference/post-analysis.ts:70-85`:

```ts
dsCoverage: dsPackages === null ? null : measureDsCoverage(context.projectDir, dsPackages),
nodeList: dsPackages === null ? undefined
  : analyzeDsCoverage({ projectDir: context.projectDir, dsPackages, includeNodes: true }).nodeList,
```

`measureDsCoverage` (`metrics/coverage.ts:119`) is `analyzeDsCoverage({projectDir, dsPackages})` plus a field projection. So this walks the module graph twice with identical inputs, and the second report already contains every field the first one projects.

To be fair on the cost: I measured it against the cached `base-ui-v1` tree (127 files) at 84ms cold / 29ms warm, memoised per pin — so this is a clarity argument, not a throughput one, and I'd rather say that than overstate it. But the shape is the problem: a reader has to notice that two calls differing by one boolean are the same measurement. Split `measureDsCoverage` into the analyzer call plus a pure `dsCoverageOf(report: DsCoverageReport): DsCoverage` projection, then baseline mode is one analysis and two projections. `deltaToBaseline` (line 164) and run mode (line 66) keep the existing two-argument helper unchanged.

### 5. `strip()` does two jobs, and the artifact is read three times per run

`analyze-results.ts:265-273` is named for removing `__run`/`__postAnalysis`, but it also reads `ds-misuse.json` off disk and merges it in. The comment explains *why* the merge happens after the cache lookup — that reasoning is sound and worth keeping — but it explains a behaviour the function's name denies.

The reads compound: `strip` runs once per group (line 378) and again over the whole set (line 387), and `main` reads the same artifacts a third time for the unjudged count (line 350). That's three reads of a file carrying one record per judged JSX node, per run, per invocation.

Read the reports once into a `Map` keyed by `runDir`, use it for both the count and the merge, and let `strip` go back to stripping.

### 6. Dead export

`lib/agentic-reference/metrics/ds-misuse/types.ts:63-66` declares `export interface NodeCensus { nodes: NodeRecord[] }`. Nothing references it — it's the only unused export in the new tree. Delete it.

### 7. Minor: the JSX tag-name dispatch is now written in three places

`node-path.ts:60` introduces `elementTag()`, whose first act is `ts.isJsxElement(e) ? e.openingElement.tagName : e.tagName`. `census.ts` imports that module and then writes the same ternary twice more — at line 112 in the `containsCountableJsx` predicate, and at line 176 where `walk` calls `count`, which this PR rewrote.

`elementTag` isn't directly reusable at those two sites (they need the `tagName` node, not its text), so the fix is either to export a `tagNameOf(element)` from `node-path.ts` alongside `elementTag` — which would then be built on it — or to drop `count`'s `tag` parameter entirely and derive it inside, since `count` already receives the element. The second is smaller and removes the ternary from `walk`.

---

**Approval bar.** (1) has to land before this is mergeable — the feature does not run on what is committed. (2) is the one I'd push hardest on structurally: the PR makes the right argument for sharing discovery and then declines to apply it to the module resolution that determines both cost and correctness. (3)–(5) are the reuse cleanups; (6) and (7) are small.

Nothing here touches the metric's design. The docs-pin decision, the model-decides-new-versus-moved tradeoff with the artifact recording the buckets, and the `node-path.ts` "do not diff path sets" warning are all well argued and well documented, and the limitations sections read like someone who went looking for the failure modes rather than around them.

agent-eval/lib/post-analysis/baseline.ts
```diff
@@ -15,24 +19,45 @@
...
 	metricsVersion?: number;
+	/**
+	 * Whether a node sidecar was written beside this file. Recorded rather than
+	 * inferred because the two legitimate reasons for a missing sidecar have to
+	 * be told apart: a module that measures no nodes (or a pin declaring no
+	 * design system) never writes one and must still hit the cache, while a
+	 * baseline that had one and lost it has to be rebuilt. Absent for legacy
+	 * files, which counts as "never had one".
+	 */
+	nodeSidecar?: boolean;
```
Sidnioulz, line 38: This is terrible naming and it should be more explicit, e.g. `hasDsCoveragePerNodeReport`.

Sidnioulz: It should prob match `has` + something to indicate the DsCoverage metric + the name used in the DS coverage script opts

agent-eval/lib/post-analysis/baseline.ts
```diff
@@ -106,43 +204,52 @@ export async function loadOrBuildBaselineAnalysis(
...
 		);
 	}
 
+	// The node list rides out to its own file: the committed baseline is meant to
+	// stay readable in a diff, and thousands of records would end that. Guarded
+	// with Array.isArray rather than a presence check, because this is the one
+	// place the module looks inside an analysis it otherwise treats as opaque —
+	// the same guard readNodeSidecar applies at the other end.
+	const { nodeList, ...analysisWithoutNodes } = analysis as Analysis & {
+		nodeList?: unknown;
```
Sidnioulz, line 206: This opt too, and all helpers should have a more explicit name.

Sidnioulz:
## Second pass on the rebase — findings + implementation plan

Re-reviewed at `4ba6cc2` (rebased onto `d20acac`). The rebase pulled in the harmonised CLI grammar from #395 (`lib/agentic-reference/selection.ts`), and the judge CLI did not come along with it. That turns two of the reuse findings from my first pass into blocking breakage, and changes what the committed-artifact fix has to be. Everything below is re-verified against `4ba6cc2` — nothing is carried over on trust.

---

## What the rebase broke

### R1 — the branch does not typecheck

```
scripts/judge-ds-misuse.ts(159,49): error TS2345:
  Argument of type 'Options' is not assignable to parameter of type 'RunSelection'.
  Type 'Options' is missing the following properties from type 'RunSelection': experiments, evals
```

`RunSelection` (`discovery.ts:23-28`) is now `{experiments: string[]; evals: string[]; since; latest}`, but `judge-ds-misuse.ts:47-52` still builds the pre-rebase `{experiment: string | null; since; latest; recompute}` and passes it straight to `selectRuns`. `scripts/judge-ds-misuse.ts` is byte-identical to its pre-rebase self — the conflict resolution updated everything around it.

(Run from a clean worktree at `4ba6cc2`. Two further errors — `@anthropic-ai/sdk` and `playwright` unresolved — are artifacts of my linking the base branch's `node_modules`; ignore those. `d20acac` typechecks clean under the same setup, so the `judge-ds-misuse` error is the branch's own.)

### R2 — analyze-results prints a rejudge command the judge rejects

`analyze-results.ts:386-388` now builds the hint from the plural flags:

```ts
'  Run: pnpm judge:ds-misuse' +
  (options.experiments.length ? ` --experiments=${options.experiments.join(',')}` : '') +
  (options.latest ? ' --latest' : '')
```

`judge-ds-misuse.ts:54-68` only accepts `--experiment=<name>`, singular, single value. So the copy-pasteable command aborts with *"Unknown argument"*. `--evals` and `--since` are dropped from the hint entirely, so even the singular form under-selects.

This is finding 3 from my first pass, promoted: it is no longer "two parsers may drift", it is "they have drifted, and the drift is on the happy path".

---

## Committed artifacts — the state moved, the problem didn't

Better than last pass: three of the five committed baselines are now at `metricsVersion: 7`. But the shape of the bug changed, and it is now self-concealing.

At `4ba6cc2`, `agent-eval/baselines/`:

| file | `metricsVersion` | `nodeSidecar` |
|---|---|---|
| `…@base-ui-v1.json` | 7 | **absent** |
| `…@droppy-70pc-v4.json` | 7 | **absent** |
| `…@droppy-v2.json` | 7 | **absent** |
| `…@droppy-70pc-v2.json` | 6 | absent |
| `…@ce507b34….json` | (none) | absent |

`baselines/ds-nodes/` is still not committed — zero files.

The three v7 files were written by commit `1c2ca97` ("bump metricsVersion to 7"), which predates `cd68b84` (sidecar) and `0394325` (pair consistency). They were never regenerated afterwards, so they carry no `nodeSidecar` field.

That matters because of `baseline.ts:171-173`:

```ts
const sidecarIntact =
  committed?.nodeSidecar !== true ||
  readNodeSidecar(baselinesDir, pin, postAnalysis.metricsVersion) !== null;
```

`absent !== true`, so `sidecarIntact` is **true**. Combined with `versionMatches`, these baselines are a permanent cache hit. I probed it directly — temp `baselinesDir` holding the committed `base-ui-v1.json`, real ref cache, a stub `postAnalysis` at version 7 whose `analyzeRun` throws if called:

```
sidecar before: ABSENT
RESULT: cache HIT — analyzeRun never called, so no sidecar is ever written
sidecar after:  ABSENT
```

So: `pnpm results:analyze` will never write the sidecars, the `'node sidecar missing'` rebuild reason can never fire for these pins, and `judge:ds-misuse` skips every run with *"no node census … Run: pnpm results:analyze --recompute"* — advice that is correct only because the plain form silently does nothing.

The "absent means never had one" rule is right for genuinely legacy files. It just also swallows this PR's own half-migrated ones. Regenerating fixes the instance; the plan below also closes the hole.

Separately, the rebase brought back five **eval-keyed** baselines that were not re-keyed — `702-rework-ui-flow/`, `703-fix-bug-flow/`, `704-fix-a11y-flow/`, `706-new-ui-scheduled-flow/`. `baselinePath()` only ever looks at `baselines/<pinSlug>.json`, so nothing reads them. They are exactly the duplication `ef1cb10` set out to delete.

---

## Your review comments

**C1 — `nodeSidecar` naming** (`baseline.ts:38`, and the `hasNodeSidecar` local at `baseline.ts:208`, plus `writeNodeSidecar`/`readNodeSidecar`/`nodeSidecarPath`/`NODE_SIDECAR_DIR`). Agreed, and your follow-up gives the rule: `has` + the metric + the name the ds-coverage CLI already uses. That opt is `--nodes` / `includeNodes`, and the payload field is `nodeList`. So `hasDsCoverageNodeList` for the flag, `dsCoverageNodeList*` for the helpers, keeping `ds-nodes/` as the directory since that name is already right. Naming the metric in the field is what stops the next reader assuming it is generic post-analysis state.

**C2 — node record format** (`node-path.ts:34`). Worth pulling apart, because two separate things are being asked of that format. The `path` string is an *identity key* — its job is to survive relocation so the sidecar and the treatment census can refer to the same node, and `basePath()` already builds it from the JSX ancestor chain, so the tree structure is in there, just flattened. What is genuinely list-shaped is the *wire* format in `context.ts:47-55`, one tab-separated line per node, chosen for token cost.

If the judge is having to reconstruct containment from those strings to decide new-versus-moved, the fix is on the wire side — nest the lines, or emit an indented tree — and it costs nothing structurally, because the AST relationships are still recoverable from the path chain. Changing `path` itself into a structured value is the expensive version: it changes the sidecar payload, forces another `metricsVersion` bump, and re-opens the `#n` disambiguation the current format pins down. I'd treat this as a follow-up measured against a real judged run rather than a change to make blind — the PR has never made a live call, so there is no evidence yet about which way the judge actually fails. Flagged as an open question in the plan, not a task.

---

## Carried over from the first pass, re-verified at `4ba6cc2`

- **Post-analysis module resolution is still not shared.** `judge-ds-misuse.ts:36` hardcodes `#lib/agentic-reference/post-analysis`; `analyze-results.ts` resolves it per experiment via `hooks.ts`. The judge has no equivalent of the no-hook skip, so it spends model calls on runs analyze-results declines to measure.
- **`messageOf` duplicated verbatim** (`judge-ds-misuse.ts:81` / `analyze-results.ts:155`); **`pinOf` near-duplicated** (`judge-ds-misuse.ts:71-79`).
- **Baseline mode analyses the same tree twice** (`post-analysis.ts:74-86`). Measured 84ms cold / 29ms warm on the 127-file `base-ui-v1` tree, memoised per pin — a clarity fix, not a throughput one. CodeRabbit flagged this independently.
- **`strip()` also reads and merges the artifact**, and `readMisuseReport` runs three times per run (`analyze-results.ts:294`, `379`, and again via `415`).
- **`NodeCensus` is dead** (`ds-misuse/types.ts:64`) — one occurrence in the tree, its own declaration.
- **The JSX tag-name ternary appears three times** — `census.ts:112`, `census.ts:185`, `node-path.ts:61`.

## Verified from CodeRabbit's pass

Not your comments, so listed separately — but I checked each against the code rather than taking the bot's word, and these four are real. Your call on whether they land here or as follow-ups.

- **`isStale` ignores the model** (`ds-misuse/index.ts:65-71`). `DsMisuseReport.model` is recorded and never compared to `JUDGE_MODEL`, so changing the judge model silently reuses old scores alongside new ones in the same table. Cheap fix, and this metric's whole premise is cross-arm comparability.
- **An empty patch widens the census to the entire project** (`tree-patch.ts:130-134` → `index.ts:83`). `censusInclude: []` means "no filter", per the contract on `DsCoverageOptions`. A run whose diff has no judgeable source file — or whose blocks were all dropped by `maxBytes` — sends every node in the tree as "treatment", at full token cost.
- **The output schema lets a node carry the wrong kind's scores** (`types.ts:103`). `required` is the five identity fields; all three score groups are optional with no dependency on `kind`. `summariseJudgement` averages by key presence, not by kind, so a `local` node returned with a `correctDsDecision` silently lands in the DS mean.
- **`strip()` merges stale reports as if current** (`analyze-results.ts:293-300`). It null-checks but never calls `isStale`, so a report from a moved guidelines pin still prints. The same check is missing from the unjudged count at `379`.

Not included: CodeRabbit's `pinSlug` collision claim needs a repo or ref containing a literal `__` to bite, which is not a real input here. The cheap half is worth taking anyway — see T9.

---

# Implementation plan

For an agent picking this up cold. Tasks are ordered so each one leaves the branch in a state that builds. `agent-eval/` is the working directory throughout; `pnpm typecheck && pnpm test && pnpm lint && pnpm format:check` is the gate after every task.

**Ground rules.** Do not make a live Anthropic API call — every judge test mocks the SDK and it stays that way. Do not change `DS_DOCS_PIN`. Do not renumber `metricsVersion` beyond what T3 specifies. Preserve the existing comment style: these modules explain *why*, and a rename that leaves the reasoning behind it stale is a regression.

### T1 — make the judge CLI use the shared selection grammar *(fixes R1 + R2)*

`scripts/judge-ds-misuse.ts`.

Replace the hand-rolled `parseArgs` (lines 54-68) with `selectionFlags(process.env)` from `#lib/agentic-reference/selection`, mirroring `analyze-results.ts:108-153`. Take `experiments`, `evals`, `since`, `latest` from the shared flag set; keep `recompute` as the judge's own switch, declared the same way analyze-results declares its own. Update the usage comment at line 21 to the plural grammar and note the `AGENTIC_REF_<FLAG>` fallback.

Then extend the rejudge hint in `analyze-results.ts:386-388` to carry `--evals` and `--since` too, so the printed command reproduces the selection that produced the warning.

**Acceptance.** `pnpm typecheck` clean. `pnpm judge:ds-misuse --experiments=a,b --evals=706 --since=2026-08-01 --latest` parses. The command printed by the unjudged warning runs without error. Add a test asserting the two CLIs resolve the same `RunSelection` from the same argv.

### T2 — share post-analysis resolution

Move `loadPostAnalysis` out of `analyze-results.ts:165+` into `lib/post-analysis/hooks.ts` beside `postAnalysisFrom`, taking the experiments directories as a parameter so both roots (`experiments/`, `.agentic-ref/experiments/`) stay caller-supplied. Have both CLIs use it.

In `judge-ds-misuse.ts`, drop the `#lib/agentic-reference/post-analysis` import and resolve the module per run's experiment. A run whose experiment names no module is `skipped`, matching analyze-results — and skipped *before* `assertApiKey`, so it never costs a call. Use the resolved module's `metricsVersion` for both `isStale` and `readNodeSidecar`.

Update the header comment on `hooks.ts` — it currently says the IO half "stays in scripts/analyze-results.ts", which stops being true here.

**Acceptance.** No `#lib/agentic-reference/*` import remains in `judge-ds-misuse.ts` except `metrics/*`, which is the metric itself. Test: a run under an experiment with no `postAnalysis` is skipped and no model call is attempted.

### T3 — regenerate the committed baselines and sidecars, and close the hole that hid it

Two parts, both needed.

**(a) Fix the detection.** In `baseline.ts`, a baseline at the current `metricsVersion` that carries no sidecar flag must not be treated as complete. Distinguish "legacy file, predates the field" from "current-version file that should have recorded it": treat an absent flag as a cache miss when `committed.metricsVersion === postAnalysis.metricsVersion`, and keep the permissive reading only for files whose version is absent or older. Extend the `reason` string so the rebuild says which case it hit. Add a test for exactly the state on disk today — v7 baseline, no `nodeSidecar`, no sidecar file — asserting a rebuild.

**(b) Regenerate.** Run `pnpm results:analyze --recompute` then `pnpm format`, and commit: all five pin-keyed baselines at `metricsVersion: 7` with an explicit `nodeSidecar` flag, plus `baselines/ds-nodes/*.json` for every pin that declares DS packages. Delete the five stale eval-keyed baselines under `702-rework-ui-flow/`, `703-fix-bug-flow/`, `704-fix-a11y-flow/`, `706-new-ui-scheduled-flow/` — `baselinePath()` cannot reach them.

**Acceptance.** From a clean clone with no `.eval-cache`, `pnpm judge:ds-misuse --latest` reaches `assertApiKey` rather than skipping on a missing census. `git status` is clean after a plain `pnpm results:analyze`. `ls baselines/` shows one file per pin and the `ds-nodes/` directory.

### T4 — rename the sidecar vocabulary *(your C1)*

`lib/post-analysis/baseline.ts` and callers (`judge-ds-misuse.ts:116`, tests).

- `CommittedBaseline.nodeSidecar` → `hasDsCoverageNodeList`
- `hasNodeSidecar` local (line 208) → same
- `writeNodeSidecar` / `readNodeSidecar` / `nodeSidecarPath` → `writeDsCoverageNodeList` / `readDsCoverageNodeList` / `dsCoverageNodeListPath`
- `CommittedNodeSidecar` → `CommittedDsCoverageNodeList`
- `NODE_SIDECAR_DIR` keeps the value `'ds-nodes'`

The field name lives in committed JSON, so land this **in the same commit as T3(b)** — regenerating and renaming separately means writing the files twice. Update the doc comments so they describe the DS-coverage node list rather than a generic sidecar.

**Acceptance.** No `nodeSidecar` identifier remains. Committed baselines carry `hasDsCoverageNodeList`.

### T5 — one census per baseline

`lib/agentic-reference/metrics/coverage.ts`: split `measureDsCoverage` into a pure projection `dsCoverageOf(report: DsCoverageReport): DsCoverage` plus the existing two-argument helper, which becomes `analyzeDsCoverage` + `dsCoverageOf`.

`post-analysis.ts:74-86`: call `analyzeDsCoverage({..., includeNodes: true})` once, feed its report to `dsCoverageOf`, and read `nodeList` off the same report. Run mode (line 66) and `deltaToBaseline` keep the existing helper.

**Acceptance.** One `analyzeDsCoverage` call in baseline mode. Baseline output byte-identical to before the change apart from key order — diff a regenerated baseline against T3's to confirm.

### T6 — reunify the small duplicates

- Move `messageOf` into `lib/utils/` and import it in both scripts.
- Export one `pinOf(result: unknown)` — `analyze-results.ts:218` is already the right shape — and have the judge call it as `pinOf(readJson(join(runDir, 'result.json')))`. Keep one copy of the "the pin the run recorded, never today's" comment, at the definition.
- Delete `NodeCensus` (`ds-misuse/types.ts:64`).

### T7 — read each misuse report once

`analyze-results.ts`. Build a `Map<string, DsMisuseReport | null>` keyed by `runDir` once, after `successfulAnalyses` is assembled. Use it for the unjudged count (379), for `strip` (294), and for the `analysis-summary.json` pass (415). Return `strip` to stripping only — pass the resolved report in, or merge at the call sites.

While here, apply the staleness check from CodeRabbit's finding: a report that fails `isStale` is not merged and counts as unjudged, with the warning saying "stale" rather than "absent" so the operator knows to pass `--recompute`.

**Acceptance.** `readMisuseReport` called at most once per run per invocation. Test: a report at a superseded `dsGuidelinesRef` does not appear in the tables and is counted as needing a rejudge.

### T8 — tighten the judge contract

Three independent changes in `lib/agentic-reference/metrics/ds-misuse/`:

1. `index.ts:65-71` — add `report.model !== JUDGE_MODEL` to `isStale`. Test the mismatch.
2. `tree-patch.ts` + `index.ts:83` — stop an empty `files` list widening the census to the whole tree. Simplest correct form: short-circuit in `judgeRun` — when `patch.files.length === 0`, skip the treatment census entirely and judge nothing, recording the run as unjudgeable rather than sending a whole-tree node list. Add a `tree-patch` test where every block exceeds `maxBytes`.
3. `types.ts:86-108` — make `JUDGE_OUTPUT_SCHEMA` an `anyOf` over the two kinds: `kind: 'ds'` requires the two DS score groups and forbids `correctLocalDecision`; `kind: 'local'` requires `correctLocalDecision` and forbids the DS pair. `additionalProperties: false` on both variants. (`anyOf` is supported; `if`/`then` is not.) Then make `summariseJudgement` partition on `kind` rather than key presence, so the two can never disagree.

### T9 — verify the sidecar belongs to the pin that asked for it

`readDsCoverageNodeList` (post-T4) reads `repo`/`ref` out of the file and ignores them. Compare against the requested pin and return `null` on mismatch, same as the version check. One condition; makes a hand-moved or slug-collided file fail closed instead of scoring a run against another tree's census.

### T10 — the tag-name ternary

`census.ts:112` and `census.ts:185` re-derive the dispatch `node-path.ts:61` already performs. Smallest fix: drop `count`'s `tag` parameter and derive it inside from the element it already receives, then export a `tagNameOf(element)` from `node-path.ts` and use it at both remaining sites, with `elementTag` built on it.

Lowest priority — take it only if T1-T9 are green.

---

### Open question, not a task

The wire format for node records (your C2). Recommend deferring until after the first live judged run, then deciding from the actual failure mode: if the judge misattributes new-versus-moved, nest the `context.ts` lines into an indented tree first — that is cheap and changes no stored artifact. Restructuring `path` itself is the expensive option and would need another `metricsVersion` bump.

### Suggested commit sequence

`T1` → `T2` → `T3(a)` → `T3(b)+T4` (one commit; the rename touches committed JSON) → `T5` → `T6` → `T7` → `T8` → `T9` → `T10`.

Full gate after each: `pnpm typecheck && pnpm test && pnpm lint && pnpm format:check`. T3(b) additionally needs a clean `git status` after a plain `pnpm results:analyze`.

agent-eval/lib/agentic-reference/metrics/ds-misuse/tree-patch.ts
```diff
@@ -0,0 +1,136 @@
...
+		bytes += size;
+		kept.push(block);
+		files.push(path);
+	}
+
+	return {
+		text: kept.join('\n'),
+		files,
+		truncated: droppedFiles > 0,
+		droppedFiles,
```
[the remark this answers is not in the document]
Sidnioulz: Skipping is the way to go if there are no changed files to analyse.

agent-eval/lib/post-analysis/baseline.ts
```diff
@@ -61,28 +84,93 @@ export interface BaselineOptions {
...
+	nodes: NodeRecord[];
+}
+
+/**
+ * The sidecar for a pin, under its own directory so `ls baselines/` still shows
+ * one file per pin. Same slug as the baseline, so the pair is obvious on disk.
+ */
+export function nodeSidecarPath(baselinesDir: string, pin: ExternalRepoPin): string {
+	return join(baselinesDir, NODE_SIDECAR_DIR, `${pinSlug(pin)}.json`);
 }
```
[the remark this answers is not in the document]
Sidnioulz: That's ok, we have basic naming conventions in place for baseline refs.

PR #401, opened 2026-08-17
Make the ds-misuse judge runnable and its cost visible

yannbf:
Found while exercising #398's judge against the 13–15 Aug runs. Four fixes, all small; the first two are the ones that block using the metric.

### `pnpm judge:ds-misuse` could not be run as documented

Two separate reasons, both silent:

- **`.env.local` was never loaded.** The script's header comment says it runs under `node --env-file-if-exists=.env.local`, and justifies it at length — without the flag the missing-key abort would name a fix that does not work. The package script was a plain `node ./scripts/…`. So it named exactly that fix: put the key in `agent-eval/.env.local`, follow the advice, get the same error. Only `export` worked.
- **`gen:agentic-ref` was never run**, unlike `results:analyze`. On a checkout without the generated experiment definitions, every agentic-reference run resolves no `postAnalysis` and is skipped — `Skipped 5 runs whose experiment carries no postAnalysis.`, exit 0, nothing judged. Correct behaviour for a deleted arm, misleading for one that was simply never generated.

### The cost was unmeasurable

The PR asks reviewers to "check `usage.cache_read_input_tokens` on the second run to confirm the ~95k-token doc prefix is being read rather than rewritten". `runJudge` discarded `message.usage` and `DsMisuseReport` had no field for it, so that check could not be performed from anything the judge stored.

The four token counts now travel in the artifact. `usage` is optional rather than a `schemaVersion` bump: artifacts written before this carry valid scores, and re-judging one to acquire a token count costs more than the count is worth.

### An exhausted balance did not stop the sweep

A missing API key stops it, on the stated reasoning that it "will fail every remaining run identically". A spent balance is identical in that respect but arrives as a 400 on whichever request happened to hit it, so it was treated as one bad run. Observed: five runs attempted, five identical `credit balance is too low` failures. `isAccountFailure` covers missing key, spent balance, auth and permission errors; genuine per-run failures (`max_tokens`, refusal, `overloaded_error`) still let the sweep continue.

### Run labels were ambiguous

One experiment can hold several result directories, each with its own `run-1`. Labels omitted the timestamp, so `agentic-ref-cc-full-opus-high/703-fix-bug-flow/run-1` named two different runs and an error line could not be traced to a directory. `Run` already carries `timestamp`; the label now uses it.

### Tests

`judgeRun`'s happy path had no test — only the null-patch path did — so the report assembly this touches was uncovered. Added, plus the usage mapping (including absent cache fields reading as `0`, not `NaN`) and the account-failure classification.

778 passing, typecheck clean, lint clean, formatted.

### Still not exercised

**No real API call has been made here either.** Both the AI Gateway and the Anthropic account were out of credit on 17 Aug, so the judge's first live run is still ahead. What these fixes do is make that run possible and make its cost legible when it happens.

### DIFF
--- a/agent-eval/lib/agentic-reference/metrics/ds-misuse/judge.ts
+++ b/agent-eval/lib/agentic-reference/metrics/ds-misuse/judge.ts
@@ -6,7 +6,7 @@
 import Anthropic from '@anthropic-ai/sdk';
 
 import type { buildJudgeRequest } from './context.ts';
-import type { JudgeResponse } from './types.ts';
+import type { JudgeResponse, JudgeUsage } from './types.ts';
 
 /** Exactly what buildJudgeRequest produces, so no cast is needed at the call site. */
 export type JudgeRequest = ReturnType<typeof buildJudgeRequest>;
@@ -25,12 +25,35 @@ export function assertApiKey(): void {
 }
 
 /**
- * Call the judge and return its structured answer.
+ * Whether a failure belongs to the account rather than to the run being judged.
+ *
+ * An exhausted balance arrives as an ordinary per-run error — the API reports it
+ * as a 400 on whichever request happened to hit it, not as a distinct status —
+ * so a caller that treats it as one bad run walks the rest of the selection into
+ * the same wall, one paid attempt at a time. Matched on the message because that
+ * is all the SDK's error carries by the time a caller sees it.
+ */
+export function isAccountFailure(message: string): boolean {
+	return (
+		message.includes('ANTHROPIC_API_KEY') ||
+		message.includes('credit balance') ||
+		message.includes('authentication_error') ||
+		message.includes('permission_error')
+	);
+}
+
+export interface JudgeResult {
+	response: JudgeResponse;
+	usage: JudgeUsage;
+}
+
+/**
+ * Call the judge and return its structured answer, with what the call cost.
  *
  * The response is schema-constrained by output_config.format, so the only
  * failures worth naming are the ones that produce no usable content at all.
  */
-export async function runJudge(request: JudgeRequest): Promise<JudgeResponse> {
+export async function runJudge(request: JudgeRequest): Promise<JudgeResult> {
 	assertApiKey();
 	const client = new Anthropic();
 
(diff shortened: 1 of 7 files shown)


agent-eval/lib/agentic-reference/metrics/ds-misuse/judge.ts
```diff
@@ -25,12 +25,35 @@ export function assertApiKey(): void {
...
 
 /**
- * Call the judge and return its structured answer.
+ * Whether a failure belongs to the account rather than to the run being judged.
+ *
+ * An exhausted balance arrives as an ordinary per-run error — the API reports it
+ * as a 400 on whichever request happened to hit it, not as a distinct status —
+ * so a caller that treats it as one bad run walks the rest of the selection into
+ * the same wall, one paid attempt at a time. Matched on the message because that
+ * is all the SDK's error carries by the time a caller sees it.
```
Sidnioulz:
```suggestion
 * Whether a failure belongs to the account rather than to the run being
 * judged. If the judge runs out of credit, there's no point in continuing
 * to try to judge. We must abort the run. So account-related failures
 * must be isolated.
```

agent-eval/lib/agentic-reference/metrics/ds-misuse/types.ts
```diff
@@ -51,6 +51,23 @@ export interface JudgeResponse {
...
 
+/**
+ * What one judge call cost, in the four token classes that are priced
+ * differently.
+ *
+ * Recorded because the guideline corpus is the whole cost story and there is no
+ * other way to see it: a request that reads the corpus from cache and one that
+ * rewrites it differ by about twentyfold, and are indistinguishable from the
+ * scores alone. `cacheRead` near zero on a second run against an unchanged pin
+ * means the prefix is being rewritten every time.
```
Sidnioulz:
```suggestion
```

agent-eval/lib/agentic-reference/metrics/ds-misuse/types.ts
```diff
@@ -71,6 +88,13 @@ export interface DsMisuseReport {
 	/** `repo@ref` of the tree the run worked on. */
 	fixtureRef: string;
 	diffTruncated: boolean;
+	/**
+	 * What the call cost. Absent on artifacts written before this was recorded,
+	 * which is why it is optional rather than a schema-version bump: the scores in
+	 * those artifacts are still valid, and re-judging to acquire a token count
+	 * would cost more than the count is worth.
+	 */
```
Sidnioulz:
```suggestion
  /** Helps understand the cost structure of the LLM judge. */
```

Sidnioulz (approved):

PR #404, opened 2026-08-20
Surface ds-misuse judgements in the report and the misuse tables

yannbf:
The judge writes per-node verdicts with reasons into each run's `ds-misuse.json`, and nothing downstream read them: the `--misuse` tables collapse each judgement to means, and the HTML report carried no misuse content at all. The reasons — the part a reader can act on — only existed on disk.

## HTML report: a DS misuse tab

`results:compare` now pools every usable run's cached judgement into a staged `misuse.json` (artifacts only, no model calls), and the report renders it as a new tab:

- Per-cell score distributions shown whole (1 · 0.5 · 0 counts with a stacked bar), never collapsed to a mean, with an em dash for a question no node received — absence of evidence, not a zero.
- Judged-vs-usable coverage per cell, highlighted when partial, with the command that judges the rest.
- A warning when artifacts in one bundle were scored against different guideline pins, since those scores are not comparable.
- Every below-perfect verdict verbatim: score, node, file:line, run, and the judge's reason.
- An unjudged bundle renders the judge commands instead of an empty table.

## CLI: reasons after the tables

`results:analyze --misuse` prints a findings section after the two mean tables: each verdict below 1 with its score, node, file:line, run, and reason. A clean set says so in one line, because silence after a findings header reads as a broken analysis rather than a clean one.

## Notes

- Reading is free everywhere: both surfaces consume the per-run cache and never call a model.
- Misuse means are deliberately not in `COMPARISON_METRICS`, so the tab is descriptive — no significance verdicts over a handful of judged nodes.
- Verified against the three judged runs currently in results/ (all score 1, so they exercise the clean path) and against fixture artifacts for the findings path; the summary-table and findings rendering are unit-tested in both surfaces.

### DIFF
--- a/agent-eval/lib/agentic-reference/comparison-metrics.ts
+++ b/agent-eval/lib/agentic-reference/comparison-metrics.ts
@@ -8,7 +8,15 @@ export interface ComparisonMetric {
 	label: string;
 	/** Dot-path into a run's analysis.json. */
 	path: string;
-	family: 'speed' | 'cost' | 'toolUse' | 'churn' | 'dsCoverage' | 'complexity' | 'diff';
+	family:
+		| 'speed'
+		| 'cost'
+		| 'toolUse'
+		| 'churn'
+		| 'dsCoverage'
+		| 'dsMisuse'
+		| 'complexity'
+		| 'diff';
 	/** log requires y > 0 (violations become reported missing values); log0 maps log(0) to 0. */
 	transform: MetricTransform;
 	direction: MetricDirection;
@@ -156,6 +164,42 @@ export const COMPARISON_METRICS: ComparisonMetric[] = [
 		label: 'DS share of component nodes Δ',
 		path: 'deltaToBaseline.coverageDelta.dsShareOfComponentNodes.delta',
 		family: 'dsCoverage',
+		transform: 'none',
+		direction: 'higher-better',
+	},
+	{
+		// The four ds-misuse metrics read per-run means over judged nodes, so
+		// every value is already normalised to [0, 1] regardless of how many
+		// nodes a diff introduced. They are null on unjudged runs (judging is a
+		// separate paid step) — the stats stage drops those rows per metric.
+		key: 'dsMisuseScore',
+		label: 'DS misuse score',
+		path: 'dsMisuse.score',
+		family: 'dsMisuse',
+		transform: 'none',
+		direction: 'higher-better',
+	},
+	{
+		key: 'dsMisuseDecision',
+		label: 'Right DS component',
+		path: 'dsMisuse.correctDsDecision',
+		family: 'dsMisuse',
+		transform: 'none',
+		direction: 'higher-better',
+	},
+	{
+		key: 'dsMisuseUsage',
+		label: 'DS usage per docs',
+		path: 'dsMisuse.correctDsUsage',
+		family: 'dsMisuse',
+		transform: 'none',
+		direction: 'higher-better',
+	},
+	{
+		key: 'dsMisuseLocalDecision',
+		label: 'Justified local components',
+		path: 'dsMisuse.correctLocalDecision',
+		family: 'dsMisuse',
 		transform: 'none',
 		direction: 'higher-better',
 	},
--- a/agent-eval/lib/agentic-reference/metrics/ds-misuse/index.test.ts
+++ b/agent-eval/lib/agentic-reference/metrics/ds-misuse/index.test.ts
@@ -85,4 +85,15 @@ describe('isStale', () => {
 			}),
 		).toBe(true);
 	});
+
+	// An LLM judge is its model: a different model applied the rubric with a
+	// different standard, so its scores must not share a table with fresh ones.
+	it('is true for a report judged by a different model', () => {
+		expect(
+			isStale(report({ model: 'claude-opus-4-7' }), {
+				dsGuidelinesRef: 'yannbf/droppy-ds@abc',
+				metricsVersion: 7,
+			}),
+		).toBe(true);
+	});
 });
--- a/agent-eval/lib/agentic-reference/metrics/ds-misuse/index.test.ts
+++ b/agent-eval/lib/agentic-reference/metrics/ds-misuse/index.test.ts
@@ -15,7 +15,7 @@ function report(overrides: Partial<DsMisuseReport> = {}): DsMisuseReport {
 	return {
 		schemaVersion: 1,
 		metricsVersion: 7,
-		judgeVersion: 1,
+		judgeVersion: DS_MISUSE_JUDGE_VERSION,
 		judgedAt: '2026-08-14T00:00:00.000Z',
 		model: 'claude-opus-4-8',
 		dsGuidelinesRef: 'yannbf/droppy-ds@abc',
@@ -108,25 +108,26 @@ describe('isStale', () => {
 	// The deterministic metricsVersion only records which census rules built
 	// the node paths; it must not invalidate a paid judge artifact.
 	it('is NOT stale when only metricsVersion differs', () => {
-		expect(isStale(report({ metricsVersion: 99 }), { dsGuidelinesRef: 'yannbf/droppy-ds@abc' })).toBe(
-			false,
-		);
+		expect(
+			isStale(report({ metricsVersion: 99 }), { dsGuidelinesRef: 'yannbf/droppy-ds@abc' }),
+		).toBe(false);
 	});
 
 	it('is true when the judge version moved', () => {
-		expect(
-			isStale(report({ judgeVersion: 0 }), { dsGuidelinesRef: 'yannbf/droppy-ds@abc' }),
-		).toBe(true);
+		expect(isStale(report({ judgeVersion: 0 }), { dsGuidelinesRef: 'yannbf/droppy-ds@abc' })).toBe(
+			true,
+		);
 	});
 
-	// An older artifact predates the judgeVersion field but was produced by
-	// judge version 1, so an absent field reads as 1 — a paid judgement must
-	// not be re-spent because the version stamp arrived after it did.
+	// An artifact predating the judgeVersion field was produced by judge
+	// version 1 and reads as such — never as automatically stale, so a paid
+	// judgement is not re-spent just because the stamp arrived after it did.
 	it('reads a report lacking judgeVersion as version 1', () => {
 		const { judgeVersion: _judgeVersion, ...legacy } = report();
-		expect(isStale(legacy as DsMisuseReport, { dsGuidelinesRef: 'yannbf/droppy-ds@abc' })).toBe(
-			DS_MISUSE_JUDGE_VERSION !== 1,
-		);
+		const stampless = legacy as DsMisuseReport;
+		expect(
+			isStale({ ...stampless, judgeVersion: 1 }, { dsGuidelinesRef: 'yannbf/droppy-ds@abc' }),
+		).toBe(isStale(stampless, { dsGuidelinesRef: 'yannbf/droppy-ds@abc' }));
 	});
 
 	it('is true for a report from an older schema', () => {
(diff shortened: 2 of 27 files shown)


yannbf:
Two additions since the original description:

- **ds-misuse is now in the comparison metric registry** — four core metrics (aggregate + the three judged questions), normalised per answer so diff size doesn't skew a run's score. Values graft from each run's cached judgement at cell-build time (stale pins excluded), flow into the dataset, estimates, verdict table, and ECDF curves, and simply stay null where judging hasn't run. The DS misuse tab now follows the workflow scope and treatment chips; significance stays inert there since per-node verdicts carry no q — the four estimate rows react to it instead.
- **The judge reports its spend** — per-run token usage and estimated dollars from list prices, plus a pass total. It also takes `--plan` now, resolved by the same helpers as `results:compare`, so the runs a plan's comparison stands on are exactly the runs you judge.

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -44,11 +47,56 @@ import { postAnalysis } from '#lib/agentic-reference/post-analysis';
...
 import { readJson } from '#lib/utils/files';
 import { isRecord } from '#lib/utils/type';
 
+import type { JudgeUsage } from '#lib/agentic-reference/metrics/ds-misuse/judge';
 import type { NodeRecord } from '#lib/agentic-reference/metrics/ds-coverage/types';
 
+// claude-opus-4-8 list prices per million tokens. The judge caches the doc
+// corpus with a 1h TTL, so the write rate is the 1h one. A moved judge model
+// means these need moving with it.
+const USD_PER_MTOK = { input: 5, cacheRead: 0.5, cacheWrite: 10, output: 25 };
```
Sidnioulz: Best move this to the `constants.ts` file

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -44,11 +47,56 @@ import { postAnalysis } from '#lib/agentic-reference/post-analysis';
...
+
+function usdOf(usage: JudgeUsage): number {
+	return (
+		(usage.inputTokens * USD_PER_MTOK.input +
+			usage.cacheReadTokens * USD_PER_MTOK.cacheRead +
+			usage.cacheWriteTokens * USD_PER_MTOK.cacheWrite +
+			usage.outputTokens * USD_PER_MTOK.output) /
+		1_000_000
+	);
+}
```
Sidnioulz: Ultimately we should have more judges, so we could also have a `judge-utils.ts` file in `metrics/`. Good way to colocate pricing info with models chosen and avoid accidents.

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -44,11 +47,56 @@ import { postAnalysis } from '#lib/agentic-reference/post-analysis';
...
+function addUsage(total: JudgeUsage, usage: JudgeUsage): void {
+	total.inputTokens += usage.inputTokens;
+	total.cacheReadTokens += usage.cacheReadTokens;
+	total.cacheWriteTokens += usage.cacheWriteTokens;
+	total.outputTokens += usage.outputTokens;
+}
+
+function tokens(n: number): string {
+	return n >= 1000 ? `${(n / 1000).toFixed(1)}k` : String(n);
+}
```
Sidnioulz: Should move to the utils that handle unit display so we also use it in the analyze script. TBH i think that function does already exist.

agent-eval/scripts/compare-results.ts
```diff
@@ -165,6 +171,17 @@ async function main() {
...
+	const misusePanel = collectMisusePanel(cells, spec, { repoRoot: resolve(ROOT, '..') });
+	writeFileSync(join(stagingDir, 'misuse.json'), JSON.stringify(misusePanel, null, 2) + '\n');
+	console.log(
+		misusePanel.judgedRuns === misusePanel.usableRuns
+			? `DS misuse: all ${misusePanel.usableRuns} usable runs judged.`
+			: `DS misuse: ${misusePanel.judgedRuns}/${misusePanel.usableRuns} usable runs judged` +
+					(misusePanel.judgedRuns === 0 ? ' — the misuse metrics will be empty' : '') +
+					`; judge the rest: pnpm judge:ds-misuse${plan === null ? '' : ` --plan ${plan.path}`}`,
+	);
+	console.log('');
```
Sidnioulz, line 186: Here I'd love to see a table of all "LLM judge analysis data:" that looks like our collected runs, with "unanalysed", "superseded", "complete", etc. and the same color coding.

Implying we need to version DS misuse results now before we ever make changes to the metric.

Sidnioulz, line 186: Here I'd love to see a table of all "LLM judge analysis data:" that looks like our collected runs, with "unanalysed", "superseded", "complete", etc. and the same color coding.

Implying we need to version DS misuse results now before we ever make changes to the metric.

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -90,13 +140,43 @@ function parseOptions(argv: string[]): Options {
...
+async function applyPlanScope(options: Options): Promise<Options> {
+	if (options.plan === undefined) {
+		return options;
+	}
+	if (options.experiments.length > 0 || options.evals.length > 0) {
+		throw new Error(
+			'--plan already names the cases and workflows; drop --experiments/--evals ' +
+				'(or unset AGENTIC_REF_EXPERIMENTS/AGENTIC_REF_EVALS).',
+		);
+	}
```
Sidnioulz: There's some overlap with `compare`, could we group up the code that handles this?

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -186,9 +272,8 @@ async function judgeOne(run: Run, options: Options): Promise<'judged' | 'reused'
 	// misconfigured environment surfaces every other problem in the same pass.
```
Sidnioulz:
```suggestion
```

agent-eval/scripts/judge-ds-misuse.ts
```diff
@@ -257,24 +367,61 @@ async function main() {
...
+		cacheReadTokens: 0,
+		cacheWriteTokens: 0,
+		outputTokens: 0,
+	};
+	const startedAt = Date.now();
+	const progress = { done: 0, total: runs.length };
+	// The pin prints once above and each run once below, so the only thing a
+	// cell contributes per run is its number — the arm and workflow become a
+	// group heading instead of repeating on every line. Lazy, so a group whose
+	// runs are all cached prints nothing at all.
```
Sidnioulz: Only the lazy comment seems useful to me

agent-eval/scripts/compare-results.ts
```diff
@@ -200,12 +217,68 @@ async function main() {
...
-	console.log(`HTML report: ${join(outDir, 'report.html')}`);
+	console.log(`HTML report: ${outStyle.bold(join(outDir, 'report.html'))}`);
+}
+
+/**
+ * The one-screen answer after the stats stage: which arms moved, which way.
+ *
+ * The counts mirror the HTML summary's better/worse tally — direction-aware,
+ * FDR-significant rows only — so the terminal says whether the report is worth
+ * opening before anyone opens it.
```
Sidnioulz:
```suggestion
 * Quick preview of the HTML summary's better/worse tally so the
 * terminal says whether the report is worth opening.
```

Sidnioulz: Fantastic idea btw

agent-eval/scripts/compare-results.ts
```diff
@@ -200,12 +217,68 @@ async function main() {
...
+		const parts = [
+			better.length > 0
+				? outStyle.tone('good', `${better.length} better`) +
+					outStyle.dim(` (${better.map(name).join(', ')})`)
+				: '',
+			worse.length > 0
+				? outStyle.tone('action', `${worse.length} worse`) +
+					outStyle.dim(` (${worse.map(name).join(', ')})`)
+				: '',
+		].filter(Boolean);
```
Sidnioulz: Wouldnt this logic be duplicated with what the report does? It should live in its own function ideally.

agent-eval/lib/agentic-reference/post-analysis.ts
```diff
@@ -281,6 +283,50 @@ function misuseOf(row: Record<string, unknown>): DsMisuseSummary | null {
...
+} as const;
+
+/**
+ * Print every below-perfect verdict with its reason.
+ *
+ * The tables above collapse the judgement to means; the reasons are the part a
+ * reader can act on, and until here they only existed inside ds-misuse.json.
+ * Perfect runs get one explicit line, because silence after a findings header
+ * reads as a broken analysis rather than a clean one.
+ */
```
Sidnioulz, line 307: I'd just print everything, perfect or not. Doesn't leave us wondering where the data is. WDYT? Was it too verbose?

agent-eval/lib/agentic-reference/post-analysis.ts
```diff
@@ -638,6 +684,8 @@ export function summarize(
 				localMean: (group.misuseLocalDecision as { mean: number | null }).mean,
 			})),
 		);
+
+		printMisuseFindings(withMisuse);
```
Sidnioulz, line 711: Struggling to see here, is this behind a `--misuse` opt?

Ideally I'd want the `analyze` command to always tell us if we need to run LLM judges, but only print the relevant LLM judged data when passed the relevant thematic option. This makes it easier to use `analyze` as a quick check in on partial data for a specific metric.

agent-eval/lib/agentic-reference/comparison-metrics.ts
```diff
@@ -159,6 +167,42 @@ export const COMPARISON_METRICS: ComparisonMetric[] = [
 		transform: 'none',
 		direction: 'higher-better',
 	},
+	{
+		// The four ds-misuse metrics read per-run means over judged nodes, so
+		// every value is already normalised to [0, 1] regardless of how many
+		// nodes a diff introduced. They are null on unjudged runs (judging is a
+		// separate paid step) — the stats stage drops those rows per metric.
```
Sidnioulz, line 174: I'm not sure if meaning before `compare` was the right decision. It's a teeny tiny loss of data, isn't it? Or do results also contain the number of judged nodes?

agent-eval/lib/agentic-reference/metrics/ds-misuse/index.ts
```diff
@@ -59,19 +61,24 @@ export function writeMisuseReport(runDir: string, report: DsMisuseReport): void
...
  */
 export function isStale(report: DsMisuseReport, current: StalenessCheck): boolean {
 	return (
 		report.schemaVersion !== DS_MISUSE_SCHEMA_VERSION ||
 		report.dsGuidelinesRef !== current.dsGuidelinesRef ||
-		report.metricsVersion !== current.metricsVersion
+		report.metricsVersion !== current.metricsVersion ||
+		report.model !== JUDGE_MODEL
 	);
 }
```
Sidnioulz, line 104: This function is unique to DS misuse, that's great.

A change to the overall `metricsVersion` so far meant "deterministic metrics must be recomputed", so in a sense it should not affect LLM judges at all (unless they reason on metrics rather than content, not the case here).

So, we don't need to compare `metricsVersion` here. Doing so would cause paid LLM judge data to be marked stale when it in fact isn't stale.

Instead, I think we should use a unique `metricsVersion` per judge. If we change the prompt, judge reference content, or any judging internals including the model, we can bump that judge's version.

agent-eval/lib/agentic-reference/metrics/ds-misuse/index.test.ts
```diff
@@ -85,4 +85,15 @@ describe('isStale', () => {
 			}),
 		).toBe(true);
 	});
+
+	// An LLM judge is its model: a different model applied the rubric with a
+	// different standard, so its scores must not share a table with fresh ones.
+	it('is true for a report judged by a different model', () => {
```
Sidnioulz, line 142: Tests need rework based on my other comment. Would love to see a test showing that the overall deterministic `metricsVersion` doesn't trigger data loss in the judge when bumping it.

